PR #18, opened 2023-05-03
Application bootstrap and  config improvements

alivinco:
- Refactored application bootrap procese and core application functions aggregated into NeatApp class.
- Small bug fixes.
- Fixed global configurations via UI and API.


cognite/neat/core/workflow/cdf_store.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         self.rules_storage_path = str(rules_storage_path)
         self.workflows_storage_type = "file"
 
-    def init_cdf_resources(self):
+    def init_cdf_resources(self, resource_type="all"):
```
doctrino, line 46: This parameter is not used. Is this intentional? If so, why is it there?

cognite/neat/core/workflow/manager.py
```diff
@@ -14,12 +14,20 @@
...
     def __init__(
         self,
-        client: CogniteClient,
-        registry_storage_type: str,
-        workflows_storage_path: Path,
-        rules_storage_path: Path,
+        client: CogniteClient = None,
+        registry_storage_type: str = "file",
+        workflows_storage_path: str = "workflows",
+        rules_storage_path: str = "rules",
```
doctrino, line 30: Path objects? That is what is sent in through the `NeatApp`

```suggestion
        workflows_storage_path: Path = Path("workflows"),
        rules_storage_path: Path = Path("rules"),
```

cognite/neat/explorer/explorer.py
```diff
@@ -54,12 +51,14 @@
...
 logging.debug(f" Config: {config.dict(exclude={'cdf_client': {'client_secret': ...}})}")
-cdf_client = get_cognite_client_from_config(config.cdf_client)
-
 
 prom_app = make_asgi_app()
 
 app = FastAPI(title="Neat")
+
+neat_app = NeatApp(config, app)
+neat_app.start()
```
doctrino, line 60: This still instantiates the CDF client when deploying the app. I think we should use `Lifespan` Events instead: https://fastapi.tiangolo.com/advanced/events/#lifespan

cognite/neat/explorer/explorer.py
```diff
@@ -75,44 +74,14 @@
...
-    exit(0)
-
-
-# signal.signal(signal.SIGINT, signal_handler)
-# signal.signal(signal.SIGTERM, signal_handler)
 @app.on_event("shutdown")
 def shutdown_event():
     logging.info("FastApi shutdown event")
-    triggers_manager.stop_scheduler_main_loop()
+    neat_app.stop()
```
doctrino, line 84: Can be replaced by a `Lifespan` event. The `shutdown` even is deprecated. https://fastapi.tiangolo.com/advanced/events/#lifespan

doctrino (changes requested):
Suggest using the Lifespan Events from FastAPI.

cognite/neat/core/workflow/manager.py
alivinco: Originally it was the same as in suggested solution but linter is complaining by stating that function should not be used  for generating default values . Not sure how to solve it in a good way.

cognite/neat/explorer/explorer.py
alivinco: Will change to Lifespan and add CdfClient as optional parameter for NeatApp that will be used if set , otherwise NeatApp will instantiate new based on configs.

cognite/neat/explorer/explorer.py
alivinco: Will replace

cognite/neat/core/workflow/cdf_store.py
alivinco: fixed

cognite/neat/core/workflow/manager.py
doctrino: Skip linter on that line maybe :)

cognite/neat/explorer/explorer.py
```diff
@@ -52,14 +53,20 @@
...
+    yield
+    logging.info("FastApi shutdown event")
+    neat_app.stop()
 
-app.mount("/metrics", prom_app)
+
+app = FastAPI(title="Neat", lifespan=lifespan)
+
+neat_app.set_http_server(app)
+neat_app.start()
```
doctrino, line 69: Shouldn't these be moved inside the lifespan function?

cognite/neat/examples/workflows/fast_graph/workflow.py
```diff
@@ -0,0 +1,311 @@
...
+        self.dataset_id: int = 0
+        self.current_step: str = None
+        self.source_graph: NeatGraphStore = None
+        self.solution_graph: NeatGraphStore = None
+        self.transformation_rules: TransformationRules = None
+        self.stop_on_error = False
+        self.count_create_assets = 0
+        self.graph_source_type = "memory"
+
+    def step_load_transformation_rules(self, flow_msg: FlowMessage = None):
```
doctrino, line 48: Not for this PR, but is it an idea to create a small library of steps that can be reused? This looks like a duplicated code from one of the other workflows.

doctrino (changes requested):
I think the `neat_app.start()` should be in the lifespan function.

cognite/neat/examples/workflows/fast_graph/workflow.py
alivinco: Yes, there is an idea of creating library and not just a library but make these standard steps available via workflow manifest/UI in addition to modules that can be used from code . It's evolution.

doctrino (approved):

PR #20, opened 2023-05-09
Workflow Approval Tests

doctrino:
Runs the default workflow and intercepts all CDF calls (assuming only assets, relationships, and labels), and dumps the resulting resources. The tests checks that it is the same resources that are created each time the test runs.

This is no change of functionality, so no bump or changelog

### DIFF
--- a/tests/api/test_workflows.py
+++ b/tests/api/test_workflows.py
@@ -1,9 +1,11 @@
 import pytest
+from cognite.client import CogniteClient
 from starlette.testclient import TestClient
 
 from cognite.neat.constants import EXAMPLE_WORKFLOWS
 from cognite.neat.core.workflow import BaseWorkflow
 from cognite.neat.core.workflow.model import WorkflowDefinition
+from cognite.neat.explorer.data_classes.rest import RunWorkflowRequest
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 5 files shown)


tests/api/test_workflows.py
```diff
@@ -29,11 +33,27 @@ def test_load_example_workflows_loaded(workflow_names: list[str], fastapi_client
...
+    )
+
+    assert response.status_code == 200
+    result = response.json()["result"]
+    assert result["error_text"] is None
+    data = {}
+    for resource_name in ["assets", "relationships", "labels"]:
+        memory: MemoryClient = getattr(cognite_client, resource_name)
+        data[resource_name] = memory.dump(ordered=True, exclude={"metadata.start_time", "metadata.update_time"})
+    data_regression.check(data, basename="default_workflow")
```
nikokaoja, line 59: Is data_regression fixture?

nikokaoja (approved):
Indeed, only assets, relationships and labels are created in this workflow

tests/api/test_workflows.py
doctrino: Yes, it comes from the `pytest` plugin `pytest-regression`. It simply takes the data dumps it to a a `yaml` file, and the next time the test runs checks that the data is the same as last time.

doctrino: It is thus very important that the data, which is a dictionary, is exactly the same as last time. Thus it is ordered, and the dynamic fields `start_time` and `update_time` are removed as these will change between each time the test runs.

nikokaoja: Aha! Now I understand! Pretty sweat !

PR #26, opened 2023-05-22
Reporting on asset update

nikokaoja:
- Detail report on asset changes
- Safety switch to raise error or skip asset in case when there is change of asset hierarchy

Breaking API change:

categorize_assets and categorize_relationships are now returning tuple (categorized_assets/_relationships, report)


CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [0.11.4] - 22-05-23
```
doctrino, line 16: Breaking change should update the minor on zero ver. That is 0.12.0

doctrino (approved):
Only thing I can see is that as this is a breaking change it should go to 0.12.0

PR #38, opened 2023-05-30
Bugfix + Improvement.

doctrino:
### Fixed
* Default `config.yaml` could not be reloaded.

### Improved
* The output messages for `load_transformation_rules_step` in all workflows by specifying which file is used.

image


nikokaoja (approved):
LGTM

PR #44, opened 2023-06-04
Added automatic config update after

alivinco:
Added automatic config update after rules file reload (configurable via UI)
Updated docs docker section in Installation guide


doctrino (approved):
Great work, tested it locally. All good.

PR #52, opened 2023-06-15
Configurable cdf client timeout

alivinco:
## [0.13.1] - 11-06-23
### Added
* Configurable cdf client timeout and max workers size. See [getting started](/getting-started.md) for details.
* Additional logic for handling `CogniteReadTimeoutError` and `CogniteDuplicatedError` during retries. This is an attempt
  to handle cases when under heavy load, requests to CDF may timeout even though the requests were processed successfully
  in eventual consistency manner.


cognite/neat/core/data_classes/config.py
```diff
@@ -51,6 +51,8 @@ class ClientConfig(BaseModel):
     client_name: str = "neat"
     base_url: str = "https://api.cognitedata.com"
     scopes: list[str] = ["project:read", "project:write"]
+    timeout: int = 60
```
doctrino, line 54: Suggest renaming this to include unit, sec, min, or hour?
```suggestion
    timeout_sec: int = 60
```

cognite/neat/core/utils.py
```diff
@@ -156,11 +158,48 @@ def retry_decorator(max_retries=2, retry_delay=3, component_name=""):
...
                     return func(*args, **kwargs)
+                except CogniteReadTimeout as e:
+                    previous_exception = e
+                    if attempt < max_retries:
+                        logging.error(
+                            f"""CogniteReadTimeout retry attempt {attempt + 1} failed for {component_name} .
+                            Retrying in {retry_delay * 2} second(s). Error:"""
+                        )
+                        logging.error(e)
+                        time.sleep(retry_delay * 2)
```
doctrino, line 174: Why are you multiplying with 2 here? This is setting the retry delay twice as long as the signature of the function indicates

cognite/neat/core/utils.py
```diff
@@ -156,11 +158,48 @@ def retry_decorator(max_retries=2, retry_delay=3, component_name=""):
...
+                            if attempt < max_retries:
+                                logging.error(
+                                    f"""CogniteDuplicatedError retry attempt {attempt + 1} failed for {component_name} .
+                                      Retrying in {retry_delay} second(s). Error:"""
+                                )
+                                logging.error(e)
+                                # incerasing delay to allow for CDF to catch up
+                                time.sleep(retry_delay * 2)
+                            else:
+                                raise e
```
doctrino, line 196: This will not work I think. If you get a duplicated error, you need to look into the error for the details on which resource was triggering it and which do not exists. There is no guarantee that all resources exist when you get a DuplicationError.

Here is an example on how to do a new call with only the timeseries that does not exists
```python
    try:
        client.time_series.create(time_series_list)
    except CogniteDuplicatedError as e:
        exists = {d["externalId"] for d in e.duplicated}
        missing_time_series = [t for t in time_series_list if t.external_id not in exists]
        client.time_series.create(missing_time_series)
```

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+        print(counter)
+        if counter == 1:
+            raise CogniteReadTimeout()
+        elif counter == 2:
+            raise CogniteDuplicatedError(duplicated=[1, 2, 3], failed=[], successful=[7, 8, 9])
+
+    try:
+        timeout_test()
+    except Exception as e:
+        assert isinstance(e, CogniteDuplicatedError)
```
doctrino, line 22:
```suggestion
    with pytest.raises(CogniteDuplicatedError)
        timeout_test()
```

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def test_retry_decorator_t1():
+    counter = 0
+
+    @retry_decorator(max_retries=4, retry_delay=0, component_name="test1")
+    def timeout_test():
+        nonlocal counter
+        counter += 1
+        print(counter)
```
doctrino, line 13:
```suggestion
```

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def test_retry_decorator_t2():
+    counter = 0
+
+    @retry_decorator(max_retries=4, retry_delay=0, component_name="test2")
+    def timeout_test():
+        nonlocal counter
+        counter += 1
+        print(counter)
```
doctrino, line 32:
```suggestion
```

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def test_retry_decorator_t3():
+    counter = 0
+
+    @retry_decorator(max_retries=4, retry_delay=0, component_name="test3")
+    def timeout_test():
+        nonlocal counter
+        counter += 1
+        print(counter)
```
doctrino, line 48:
```suggestion
```

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+    @retry_decorator(max_retries=4, retry_delay=0, component_name="test4")
+    def timeout_test():
+        nonlocal counter
+        counter += 1
+        raise CogniteReadTimeout()
+
+    try:
+        timeout_test()
+    except Exception as e:
+        assert isinstance(e, CogniteReadTimeout)
```
doctrino, line 69:
```suggestion
    with pytest.raises(CogniteReadTimeout)
        timeout_test()
```

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+        nonlocal counter
+        counter += 1
+        print(counter)
+        raise Exception("test5")
+
+    try:
+        timeout_test()
+    except Exception as e:
+        assert isinstance(e, Exception)
+        assert counter == 5
```
doctrino, line 86:
```suggestion
    with pytest.raises(Exception)
        timeout_test()
```

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_retry_decorator_t5():
+    counter = 0
+
+    @retry_decorator(max_retries=4, retry_delay=0, component_name="test5")
+    def timeout_test():
+        nonlocal counter
+        counter += 1
+        print(counter)
+        raise Exception("test5")
```
doctrino, line 80: This test is not testing what I think you are trying to test. The CogniteDuplicatedError and CogniteReadTimeout both inherit from Exception, thus if either of those are raised this test will pass, which I think is not what you want.

doctrino (commented):
A few things I think needs to be addressed.

cognite/neat/core/data_classes/config.py
alivinco: Reusing the same naming from CDF SDK . In UI and docs it's mentioned that it's in seconds.

tests/test_utils.py
alivinco: The test is testing that retry decorator works with any exception. The test is not related to latest fix , it's missing test for basic decorator functionality.

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+        print(counter)
+        if counter == 1:
+            raise CogniteReadTimeout()
+        elif counter == 2:
+            raise CogniteDuplicatedError(duplicated=[1, 2, 3], failed=[], successful=[])
+
+    timeout_test()
+
+
+def test_retry_decorator_t3():
```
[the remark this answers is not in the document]
alivinco: The intention was to test with >1 number of Timeout errors . But I see that first test might be redundant , will remove that.

doctrino (approved):

PR #56, opened 2023-07-07
Updated transformation rules

nikokaoja:
- Added additional validators to comply with CDF DM
- Added new fields to handle request for having both entity ids and entity names
- Added new fields to capture necessary information to resolve sheets as FDM


cognite/neat/core/data_classes/transformation_rules.py
```diff
@@ -37,39 +41,59 @@ class URL(BaseModel):
...
+        allow_population_by_field_name = True
+        anystr_strip_whitespace = True
+
+
+class_id_compliance_regex = r"^([a-zA-Z]+[a-zA-Z0-9]+[._-]{0,1}[a-zA-Z0-9]+)+$"
+
 
 class Class(Resource):
+    class_id: str = Field(alias="Class", min_length=1, max_length=255)
+    class_name: str = Field(alias="Name", min_length=1, max_length=255, default=None)
```
doctrino, line 94: You can add this to the top of the file 
```python
from pydantic import constr

ExternalId = constr(min_length=1, max_length=255)
```
Then

```suggestion
    class_id: ExternalId = Field(alias="Class", )
    class_name: ExternalId = Field(alias="Name", default=None)
```

doctrino (approved):
Just a small suggestion

PR #69, opened 2023-07-20
Refactor rules2graphql

nikokaoja:
### Changed

- Switch to using `jinja2` template engine instead of `graphql-core` for generation of GraphQL schema

### Added

- Downloading rules from private github repository


cognite/neat/core/rules/exporter/rules2graphql.py
```diff
@@ -1,21 +1,99 @@
...
     are_properties_redefined,
 )
-from cognite.neat.core.rules.models import DATA_TYPE_MAPPING, Property, TransformationRules
+
+from cognite.neat.core.rules.analysis import get_class_property_pairs
+from cognite.neat.core.rules.models import DATA_TYPE_MAPPING, TransformationRules
 from cognite.neat.core.utils.utils import generate_exception_report
 
 
+TYPE = (
```
doctrino, line 13: I suggest making all these constants private, as they should not be used outside of this module. 
```suggestion
_TYPE = (
```

cognite/neat/core/rules/exporter/rules2graphql.py
```diff
@@ -1,21 +1,99 @@
...
+            "type": TYPE,
+            "field_header": FIELD_HEADER,
+            "field": FIELD,
+            "value_type": FIELD_VALUE_TYPE,
+            "edge_occurrence": EDGE_OCCURRENCE,
+            "attribute_occurrence": ATTRIBUTE_OCCURRENCE,
+        }
+    ),
+    cache_size=1000,
+).get_template("type")
```
doctrino, line 91: I think we should avoid executing code on import other than constants. So I suggest that this could be a class lokking something like this:

```suggestion
class Rules2GraphQL:
    def __init__(self):
        self.template = Environment(
            loader=DictLoader(
                {
                    "type_header": TYPE_HEADER,
                    "type": TYPE,
                    "field_header": FIELD_HEADER,
                    "field": FIELD,
                    "value_type": FIELD_VALUE_TYPE,
                    "edge_occurrence": EDGE_OCCURRENCE,
                    "attribute_occurrence": ATTRIBUTE_OCCURRENCE,
                }
            ),
            cache_size=1000,
        ).get_template("type")

    def to_schema(self,
                  transformation_rules: TransformationRules,
                  header: bool = False
                  ) -> str:
              ...
```

cognite/neat/core/rules/exporter/rules2graphql.py
```diff
@@ -1,21 +1,99 @@
...
+
+
 def rules2graphql_schema(
     transformation_rules: TransformationRules,
+    header: bool = False,
```
doctrino, line 96: This is missing from the docstring.

doctrino (commented):
Overall I am impressed, very elegant implementation. 

In addition, you need to at `jinja2` as a primary dependency. Or optional with defined extra.

cognite/neat/core/rules/exporter/rules2graphql.py
doctrino: In the case we decide to make `jinja` an optional dependency, then we can do the local import in the init method.

cognite/neat/core/rules/exporter/rules2graphql.py
nikokaoja: Agree, implemented

cognite/neat/core/rules/exporter/rules2graphql.py
nikokaoja: Made pydantic class, with @property `schema` , and made local imports

cognite/neat/core/rules/exporter/rules2graphql.py
nikokaoja: now this is field `verbose` under `GraphQLSchema` class

doctrino (approved):
Nice, good job

PR #70, opened 2023-07-21
Standard step components (poc)

alivinco:


cognite/neat/core/workflow/step_model.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Step(ABC):
+    def __init__(self, metrics):
+        self.log: bool = False
+        self.metrics = metrics
+
+    def set_global_configs(self, cdf_client: CogniteClient, data_store_path: str, rules_storage_path: str):
+        self.cdf_client = cdf_client
+        self.data_store_path = data_store_path
+        self.rules_storage_path = rules_storage_path
```
doctrino, line 31: I not sure we need a global config. We can keep it for now, but I prefer that we try to not use it, as it adds extra requirement on the user of this functionality since then there are suddenly some extra data which they have to know about in addition to what is passed in the run method.

doctrino (approved):

PR #79, opened 2023-07-25
Refactor rules parser and test/package reorg

nikokaoja:
- Merge rules loaders/parsers into parsers
- Reorg package
- Reorg tests to represent package org


doctrino (approved):
Looks good

PR #81, opened 2023-07-26
Refactor workflows

nikokaoja:
First pass of cleaning up `workflows` module


tests/test_loader.py
```diff
@@ -12,8 +12,14 @@ def test_load_excel(transformation_rules: TransformationRules):
 
 def test_copy_examples_to_directory(tmp_path: Path):
     target_path = tmp_path / "data"
-
+    print("tmp_path")
```
doctrino, line 15:
```suggestion
```

Dockerfile
```diff
@@ -1,13 +1,41 @@
...
 COPY cognite/neat /app/cognite/neat
 
 RUN mkdir -p /app/data \
-    && cp /app/cognite/neat/workflows/examples/. /app/data -r \
+    && mkdir -p /app/data/workflows \
+    && mkdir -p /app/data/rules \
+    && mkdir -p /app/data/source-graphs \
+    && cp /app/cognite/neat/workflows/examples/. /app/data/workflows -r \
+    && cp /app/cognite/neat/rules/examples/. /app/data/rules -r \
+    && cp /app/cognite/neat/graph/examples/. /app/data/source-graphs -r \
```
doctrino, line 38: Are these necessary?  I think the explorer auto copies over when is starts up.

doctrino (approved):
Primarily moving around files?

Dockerfile
nikokaoja: I will test whether this really is needed, I guess not, it is probably left over from where we did not have copy method

PR #99, opened 2023-08-24
Feature create fdm instances

nikokaoja:
### Added

- Generation of DM instances
- `DMSDataModelFromRules`, `GenerateCDFNodesAndEdgesFromGraph`, `UploadCDFNodes` and `UploadCDFEdges` added to step libary

### Improved
- Handling of generation of pydantic model instances in case of incomplete graph instances


cognite/neat/graph/loaders/rdf_to_dms.py
```diff
@@ -0,0 +1,187 @@
...
+        message: Message to logged. Defaults to "Upload".
+        max_retries: Maximum times to retry the upload. Defaults to 1.
+        retry_delay: Time delay before retrying the upload. Defaults to 3.
+    """
+    total = len(nodes_or_edges)
+    counter = 0
+
+    if push_type not in ["nodes", "edges"]:
+        logging.info(f"push_type {push_type} not supported")
+        raise ValueError(f"push_type {push_type} not supported")
```
doctrino, line 164: Duplicated parameter, risks the user of this function sending in edges when they are supplying nodes. 
```suggestion
    message: str = "Upload",
    max_retries: int = 1,
    retry_delay: int = 3,
):
    """Uploads nodes or edges in batches

    Args:
        client: Instance of CogniteClient
        nodes_or_edges: List of nodes or edges
        batch_size: Size of batch. Defaults to 1000.
        push_type: Type of push, either "nodes" or "edges". Defaults to "nodes".
        message: Message to logged. Defaults to "Upload".
        max_retries: Maximum times to retry the upload. Defaults to 1.
        retry_delay: Time delay before retrying the upload. Defaults to 3.
    """
    total = len(nodes_or_edges)
    counter = 0

    if nodes_or_edges and isinstance(nodes_or_edges[0], NodeApply):
        push_type = "nodes"
    elif nodes_or_edges and isinstance(nodes_or_edges[0], EdgeApply):
        push_type = "edges"
    else:
        raise ValueError("nodes_or_edges must be a list of NodeApply or EdgeApply objects")
```

cognite/neat/graph/loaders/rdf_to_dms.py
```diff
@@ -0,0 +1,187 @@
...
+        @retry_decorator(max_retries=max_retries, retry_delay=retry_delay, component_name=f"microbatch-{push_type}")
+        def upsert_nodes_or_edges(batch):
+            if push_type == "nodes":
+                client.data_modeling.instances.apply(
+                    nodes=nodes_or_edges, auto_create_start_nodes=True, auto_create_end_nodes=True
+                )
+            elif push_type == "edges":
+                client.data_modeling.instances.apply(
+                    edges=nodes_or_edges, auto_create_start_nodes=True, auto_create_end_nodes=True
+                )
```
doctrino, line 179: 1. All nodes/edges are  uploaded every time. Fixed by using correct variable
2.  In addition, suggest renaming batch to avoid shadowing outer name. 
3.  Remove `auto_create_start_nodes=True, auto_create_end_nodes=True` for node as they have no effect. 

```suggestion
        def upsert_nodes_or_edges(upload_batch):
            if push_type == "nodes":
                client.data_modeling.instances.apply(nodes=upload_batch)
            elif push_type == "edges":
                client.data_modeling.instances.apply(
                    edges=upload_batch, auto_create_start_nodes=True, auto_create_end_nodes=True
                )
```

cognite/neat/graph/extractors/graph_sheet_to_graph.py
```diff
@@ -102,28 +102,58 @@ def sheet2triples(
...
+                            )
+                            for v in value.split(separator)
+                        )
+                    else:
+                        triples.append(
+                            (
+                                instance_namespace[row.identifier],
+                                model_namespace[property_],
+                                instance_namespace[value.strip()],
+                            )
```
doctrino, line 125: The nested conditionals becomes duplicated. Suggest rewriting this the conditionals in the loop as follows.
Note I renamed `property_` -> `property_name`. And use the `property_` variable for the property.
```python
            # iterate over sheet rows properties
            for property_name, value in row.to_dict().items():
                # Setting RDF type of the instance
                if property_name == "identifier":
                    triples.append(
                        (
                            instance_namespace[row.identifier],
                            RDF.type,
                            model_namespace[sheet_name],
                        )
                    )
                    continue
                elif not value:
                    continue

                property_ = class_property_pairs[sheet_name][property_name]
                is_one_to_many = (property_.max_count or 1) > 1 and separator
                values = value.split(separator) if is_one_to_many else [value]

                # Adding object properties
                if property_.property_type == "ObjectProperty":
                    triples.extend(
                        (
                            instance_namespace[row.identifier],
                            model_namespace[property_name],
                            instance_namespace[v.strip()],
                        )
                        for v in values
                    )
                # Adding data properties
                elif property_.property_type == "DatatypeProperty":
                    triples.extend(
                        (
                            instance_namespace[row.identifier],
                            model_namespace[property_name],
                            Literal(
                                v.strip(),
                                datatype=XSD[property_.expected_value_type],
                            ),
                        )
                        for v in values
                    )
```

doctrino: Consider what should happen if `not value` or whether there should be an `else` at the end of the loop. Do we expect other types than `ObjectProperty` or `DatatypeProperty`. If not this should raise an exception.

cognite/neat/rules/exporter/rules2pydantic_models.py
```diff
@@ -191,6 +206,9 @@ def _dictionary_to_pydantic_model(
     if methods:
         for method in methods:
             setattr(model, method.__name__, method)
+    if property_attributes:
+        for property_attribute in property_attributes:
+            setattr(model, property_attribute.fget.__name__, property_attribute)
```
doctrino, line 211: 🏌
```suggestion
    for property_attribute in (property_attributes or []):
        setattr(model, property_attribute.fget.__name__, property_attribute)
```

cognite/neat/rules/exporter/rules2pydantic_models.py
```diff
@@ -199,33 +217,79 @@ def _dictionary_to_pydantic_model(
     return model
 
 
+@property
+def attributes(self) -> list[str]:
+    return [
+        field
+        for field in self.model_fields_set
+        if self.model_fields[field].json_schema_extra
+        and self.model_fields[field].json_schema_extra.get("property_type", "") == "NodeAttribute"
```
doctrino, line 226: 🏌
```suggestion
        if (schema := self.model_fields[field].json_schema_extra) and schema.get("property_type") == "NodeAttribute"
```

cognite/neat/rules/exporter/rules2pydantic_models.py
```diff
@@ -199,33 +217,79 @@ def _dictionary_to_pydantic_model(
...
+    ]
+
+
+@property
+def edges_one_to_one(self) -> list[str]:
+    return [
+        field
+        for field in self.model_fields_set
+        if self.model_fields[field].json_schema_extra
+        and self.model_fields[field].json_schema_extra.get("property_type", "") == "EdgeOneToOne"
```
doctrino, line 236:
```suggestion
        if (schema := self.model_fields[field].json_schema_extra) and schema.get("property_type") == "EdgeOneToOne"
```

cognite/neat/rules/exporter/rules2pydantic_models.py
```diff
@@ -199,33 +217,79 @@ def _dictionary_to_pydantic_model(
...
+    ]
+
+
+@property
+def edges_one_to_many(self) -> list[str]:
+    return [
+        field
+        for field in self.model_fields_set
+        if self.model_fields[field].json_schema_extra
+        and self.model_fields[field].json_schema_extra.get("property_type", "") == "EdgeOneToMany"
```
doctrino, line 246:
```suggestion
        if (schema := self.model_fields[field].json_schema_extra) and schema.get("property_type") == "EdgeOneToMany"
```

cognite/neat/workflows/steps/lib/graph_loader.py
```diff
@@ -51,6 +61,75 @@ def run(self, rules: RulesData, cdf_client: CogniteClient) -> None:
...
+
+    configuration_templates: ClassVar[list[WorkflowConfigItem]] = [
+        WorkflowConfigItem(
+            name="nodes_and_edges_generation.graph_name",
+            value="source",
+            label=("The name of the graph to be used for matching." " Supported options : source, solution"),
+        ),
+    ]
+
+    def run(self, rules: RulesData, graph: SourceGraph | SolutionGraph) -> (FlowMessage, Nodes, Edges):
```
doctrino, line 80: I think you need to do this
```suggestion
    def run(self, rules: RulesData, graph: SourceGraph | SolutionGraph) -> tuple[FlowMessage, Nodes, Edges]:
```

cognite/neat/workflows/steps/step_model.py
```diff
@@ -7,6 +8,16 @@
 from cognite.neat.workflows.model import WorkflowConfigItem, WorkflowConfigs
 
 
+class StepCategory(StrEnum):
```
doctrino, line 11: Same comment as before, suggest getting the category from the module name instead to avoid redundant informration.

doctrino (changes requested):
A few things here and there. The only important fix is the micro batch function that updates the all nodes/edges for each iteration instead of just a batch.

cognite/neat/workflows/steps/step_model.py
nikokaoja: This is unused, forgot to remove it.

cognite/neat/workflows/steps/lib/graph_loader.py
nikokaoja: The definition of output is correct, 

If one adds `tuple` is incorrect here is how it is rendered in UI:
Screenshot 2023-08-28 at 08 03 14

Alt, if one use the existing definition of output (i.e. without `tuple`):

Screenshot 2023-08-28 at 08 07 22

cognite/neat/graph/extractors/graph_sheet_to_graph.py
nikokaoja: added exception

PR #105, opened 2023-08-30
New workflow configurations and auth_token

alivinco:
### Changed

- New way of configuring workflows steps . Now steps are configured individually and not as a part of workflow manifest.
- Added access_token authentication for Cognite client. If client_id is not set in config.yaml, NEAT will use client_secret as access_token.
- Multiple UI usability improvements and bug fixes.

### Added

- Added SimpleContextualization step . The step can be used to create links between nodes in a graph either by using regex or exact match between source and target properties.
- Added single store configuration step. Now solution and source graph stores can be configured individually.


cognite/neat/utils/utils.py
```diff
@@ -27,6 +27,11 @@ def get_cognite_client_from_config(config: ServiceCogniteClient) -> CogniteClien
     return _get_cognite_client(config, credentials)
 
 
+def get_cognite_client_from_token(config: ServiceCogniteClient) -> CogniteClient:
+    credentials = Token(config.client_secret)
```
doctrino, line 31: Ideally we should change the config to explicitly state which how the client should be created.

cognite/neat/workflows/steps/lib/graph_contextualization.py
```diff
@@ -0,0 +1,144 @@
...
+            target_class=self.configs["target_class"],
+            target_property=self.configs["target_property"],
+            relationship_name=self.configs["relationship_name"],
+            link_direction=self.configs["link_direction"],
+            matching_method=self.configs["matching_method"],
+        )
+        output_text = f"Matcher has created {new_links_counter} links"
+        return FlowMessage(output_text=output_text)
+
+    def simple_entity_matcher(
```
doctrino, line 72: I suggest having this methods somewhere in the graph module. I think steps should just be small wrappers around the core functionality.

alivinco: Good catch , at least that is true for NEAT provided steps.

PR #110, opened 2023-09-05
Fix none columns and erroneous dms upload step

### LINKED ISSUE #68: New relationships do not have a start_time
**Describe the bug**
Relationships created by Neat do not have any timestamp in the `start_time` field in CDF.

**To Reproduce**
Run any workflow that procudes relationships, when inspecting these via API/SDK they will have start_time = None.

nikokaoja:
This one deals with:
- issue when excel sheets contain columns which names are non-string, especially None
- dms upload step had error
- closes #68


cognite/neat/rules/parser.py
```diff
@@ -310,6 +310,20 @@ def has_instances_mandatory_columns(cls, v):
             raise exceptions.InstancesSheetMissingMandatoryColumns(missing_columns).to_pydantic_custom_error()
         return v
 
+    @staticmethod
+    def drop_non_string_columns(df: pd.DataFrame) -> pd.DataFrame:
+        """Drop non string columns as this can cause issue when loading rules
```
doctrino, line 315:
```suggestion
        """Drop non-string columns as this can cause issue when loading rules
```

cognite/neat/workflows/steps/lib/graph_loader.py
```diff
@@ -321,10 +321,10 @@ class GenerateCDFRelationshipsFromGraph(Step[CategorizedRelationships]):
 
     def run(  # type: ignore[override]
         self, rules: RulesData, cdf_client: CogniteClient, solution_graph: SolutionGraph
-    ) -> (FlowMessage, CategorizedRelationships):
+    ) -> tuple[FlowMessage, CategorizedRelationships]:
```
doctrino, line 324: Did we agree that this should be? I think Aleks used that in the frontend. 
```suggestion
    ) -> (FlowMessage, CategorizedRelationships):
```

cognite/neat/workflows/steps/lib/graph_loader.py
```diff
@@ -321,10 +321,10 @@ class GenerateCDFRelationshipsFromGraph(Step[CategorizedRelationships]):
...
         # create, categorize and upload relationships
         rdf_relationships = rdf2relationships(
-            solution_graph.graph.get_graph(),
+            solution_graph.graph,
```
doctrino, line 327: Bugfix?  I considered doing this in my PR, but I did not want to change behavior when I only did `mypy`

cognite/neat/rules/parser.py
```diff
@@ -246,14 +246,14 @@ def from_raw_dataframes(cls, raw_dfs: dict[str, pd.DataFrame]) -> "RawTables":
 
         tables_dict = {
             Tables.metadata: raw_dfs[Tables.metadata],
```
doctrino, line 248: Why not on the metadata table?

doctrino (approved):
Just a few questions.

cognite/neat/rules/parser.py
nikokaoja: you do not have column names, `metadata` works more as key value pairs, other sheets are actually tables

cognite/neat/workflows/steps/lib/graph_loader.py
nikokaoja: we did, probably I made mistake when merging main changes

cognite/neat/workflows/steps/lib/graph_loader.py
nikokaoja: solution_graph.graph -> provides graph_store, and that is what `rdf2relationships` expects as input

PR #115, opened 2023-09-09
MyPy on neat/graph

doctrino:
Before
```
Found 676 errors in 45 files (checked 1206 source files)
```

After
```
Found 446 errors in 32 files (checked 1206 source files)
```

And after deleting the inheritance based workflows
```
Found 359 errors in 28 files (checked 1201 source files)
```


cognite/neat/rules/to_rdf_path.py
```diff
@@ -142,11 +142,10 @@ def from_string(cls, raw: str, **kwargs) -> Self:
 
 
 class Traversal(BaseModel):
-    pass
```
nikokaoja, line 145: I hope this works, just hope we do not relive January nightmares

cognite/neat/utils/utils.py
```diff
@@ -115,17 +115,14 @@ def remove_namespace(
...
     """Removes namespace from URI
 
-    Parameters
-    ----------
-    URI : URIRef | str
-        URI of an entity
-    special_separator : str
-        Special separator to use instead of # or / if present in URI
-        Set by default to "#_" which covers special client use case
+    Args
```
nikokaoja, line 118: great!

cognite/neat/workflows/inheritance_based/graph2assets_relationships.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, name: str, client: CogniteClient):
         self.transformation_rules: TransformationRules = None
         self.graph_source_type = "memory"
```
nikokaoja, line 47: @alivinco should we deprecate inheritance_based workflows ? and remove this module?

nikokaoja (commented):
Some comments, mainly worried with updates of `_to_construct_triples` method, and changes in `to_rdf_path` `Traversal` class

cognite/neat/graph/transformations/query_generator/sparql.py
```diff
@@ -279,7 +280,7 @@ def build_sparql_query(
 
     if isinstance(traversal, AllProperties):
         query = _generate_all_properties_query_statement(triples[0].subject)
-    elif isinstance(traversal, AllReferences):
+    elif isinstance(traversal, AllReferences) and isinstance(triples[0].object, str):
```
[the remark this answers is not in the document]
doctrino: mypy

cognite/neat/rules/to_rdf_path.py
doctrino: I checked it fairly carefully. so think it should be fine

doctrino: I did an extra check of all the usage, and the fact that `traversal.class_` was already assumed almost all the places it was used. In addition, this is adding variable to a class (all subclasses are unchanged since they already have it). So I think it is very safe to do this change.

cognite/neat/graph/transformations/query_generator/sparql.py
```diff
@@ -450,68 +451,61 @@ def _to_construct_triples(
     tuple[list[Triple],list[Triple]]
         Tuple of triples that define graph template and graph pattern parts of CONSTRUCT query
     """
```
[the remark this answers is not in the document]
doctrino: The reason for the change is that the last implementation made a lot assumptions on types which made mypy go crazy. 

Advantage of the new methods is more performant with fewer loops, and no need to do the extra copy, meaning some saved memory. But my software brain also find it easier to follow. 

However, as I lack the deep domain experience we can revert it, but then I think we need to let `mypy` skip this function.

cognite/neat/rules/analysis.py
```diff
@@ -68,8 +68,7 @@ def to_class_property_pairs(
                 # TODO: use appropriate Warning class from _exceptions.py
                 # if missing make one !
                 warnings.warn(
```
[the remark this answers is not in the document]
doctrino: Lets save that for another PR replacing all `warnings.warn` with NeatWarnings.

PR #123, opened 2023-09-25
Feature xid class prefix

msolvik:
Updates to generate FDM from RDF:

1) Function to add class name as prefix to instance external id (to ensure uniqueness if same external id is used across classes) 
2) Reduce default batch size when uploading nodes to 1000


alivinco:
The PR if failing on mypy checks but otherwise it look fine

alivinco (approved):

PR #131, opened 2023-10-06
Rewrite Tutorial Part 1

doctrino:
Simplification of tutorial part 1 to lower the bar to get started. 
Main change
* Moved the create your own sheet into an optional section marked as advanced. 

In addition, introduced container classes for `Class` and `Property` in `TransformationRules`, this allows for some useful utility methods. The first I implemented was `.to_pandas()` and `_repr_html_` which are useful in the tutorial to illustrate the parsed rules.

Ended up skipping the test on the notebook as there is too much variability in the output, and the `pytest-notebook` package regex replacement does not work as advertised. It is a trade-off, but I think having more output in the tutorial makes the tutorial better, which is more important than the test.


doctrino:
Included in #133

PR #136, opened 2023-10-10
Feature xid class prefix

msolvik:
Update to edges creation when add_class_prefix is set to true. This is to ensure uniqueness if same external_id is used across classes. This can happen as the external_id of a node is created based on the URI. If there is a node with the URI ".../substation/1234" and another node with URI "../station/1234" the external_id is not unique, as the namespace stripping function takes the string after the last "/" as external_id. To avoid this, an option is added to the graph_loader step where the user can set add_class_prefix to true, so the external_ids of the example above are "SubStation_1234" and "Station_1234". 

This functionality is already implemented for nodes. This PR implements the same logic for edges.


cognite/neat/rules/exporter/rules2pydantic_models.py
```diff
@@ -495,21 +495,47 @@ def is_external_id_valid(external_id: str) -> bool:
...
-                external_id=f"{self.external_id}-{end_node_id}",
-                type=(data_model.space, edge_type_id),
-                start_node=(data_model.space, self.external_id),
-                end_node=(
-                    data_model.space,
-                    end_node_id,
-                ),
+        if add_class_prefix:
+            edges_list = []
+            dm_view = data_model.views[self.__class__.__name__]
```
doctrino, line 500: Just my personal preference 
```suggestion
            dm_view = data_model.views[type(self).__name__]
```

cognite/neat/rules/exporter/rules2pydantic_models.py
```diff
@@ -495,21 +495,47 @@ def is_external_id_valid(external_id: str) -> bool:
...
+            if dm_view.properties:
+                for end_node in self.__getattribute__(edge_one_to_many):
```
doctrino, line 502: Lets avoid the __gettribute__ dunder method and instead use the built in. 
```suggestion
                for end_node in getattr(self, edge_one_to_many):
```

cognite/neat/rules/exporter/rules2pydantic_models.py
```diff
@@ -495,21 +495,47 @@ def is_external_id_valid(external_id: str) -> bool:
...
+                    if is_external_id_valid(end_node):
+                        mapped_instance = dm_view.properties[edge_one_to_many]
+                        if isinstance(mapped_instance, SingleHopConnectionDefinitionApply):
+                            object_view = mapped_instance.source
+                            if object_view:
```
doctrino, line 507: All these nested if makes the code very hard to read.

cognite/neat/rules/exporter/rules2pydantic_models.py
```diff
@@ -495,21 +495,47 @@ def is_external_id_valid(external_id: str) -> bool:
...
+                        end_node_id,
+                    ),
+                )
+                for end_node_id in self.__getattribute__(edge_one_to_many)
+                if is_external_id_valid(end_node_id)
             )
-            for end_node_id in self.__getattribute__(edge_one_to_many)
-            if is_external_id_valid(end_node_id)
-        )
     return edges
```
doctrino, line 539: Suggest rewriting this entire implementation to something like this (you have to verify the implementation)

```python
    class_name = type(self).__name__

    for edge_one_to_many in self.edges_one_to_many:
        edge_type_id = f"{class_name}.{edge_one_to_many}"
        edge_list = []
        for end_node_id in getattr(self, edge_one_to_many):
            if not is_external_id_valid(end_node_id):
                continue
            end_node_external_id = end_node_id
            if add_class_prefix:
                end_node = _get_end_node(data_model[class_name], edge_one_to_many)
                if end_node is None:
                    continue
                end_node_external_id = add_class_prefix_to_xid(end_node, end_node_id)

            edge = EdgeApply(
                space=data_model.space,
                external_id=f"{self.external_id}-{end_node_external_id}",
                type=(data_model.space, edge_type_id),
                start_node=(data_model.space, self.external_id),
                end_node=(data_model.space, end_node_external_id),
            )
            edge_list.append(edge)

    return edges


def _get_end_node(view: ViewApply, edge_one_to_many: str) -> str | None:
    mapping = view.properties[edge_one_to_many]
    if isinstance(view, SingleHopConnectionDefinitionApply) and mapping.source:
        return mapping.source.external_id
    return None
```

cognite/neat/rules/exporter/rules2pydantic_models.py
doctrino, line 507: Another question is what happens if any of these if are false? Is that a problem that the edge will be skipped? Could this be a silent error when you call this method and expect that it works, and then it skips a whole lot of edges?

cognite/neat/rules/exporter/rules2pydantic_models.py
msolvik: Ok, self.__class__.__name__ is used in the entire file, so I just kept the same style. I might change the entire file then, to keep it consistent.

cognite/neat/rules/exporter/rules2pydantic_models.py
msolvik: Same comment as above

nikokaoja (approved):

PR #137, opened 2023-10-11
JSON Importer

doctrino:
Started on the XML importer but realized that the customer was deviating from the spec a lot, so it is easier just to sample a subset of the data.

I did not replace the workflow step as that contains a feature for making FDM compliant names. This needs to be part of the exporter and not importer, so if I replace that step it will not be equivalent.


cognite/neat/rules/importer/xml2tables.py
```diff
@@ -0,0 +1,22 @@
...
+        xml_directory: Path to directory containing XML files.
+    """
+
+    def __init__(self, xml_directory: Path):
+        self.xml_directory = xml_directory
+        super().__init__(spreadsheet_path=xml_directory / "transformation_rules.xlsx")
+
+    def to_tables(self) -> RawTables:
+        raise NotImplementedError
+        # raise NotImplementedError
```
doctrino, line 22:
```suggestion
```

PR #141, opened 2023-10-21
DMS Importer

doctrino:
There are a few issues with parent assets and very core CDF data model columns in the Rules. We need a discussion on whether these can be simplified.


nikokaoja (approved):
LGTM

PR #144, opened 2023-10-24
fix for not accepting changes of parent asset

msolvik:
allowing changes of parent asset external id


nikokaoja (commented):
Could you please write test for this?
Does this handles situation when root asset is changed ?

nikokaoja (approved):
LGTM

PR #161, opened 2023-11-08
from instance to class property methods

nikokaoja:


nikokaoja:
## [0.39.1] - 08-11-23
## Fixed
 - Changed `attributes`, `edges_one_to_one`, `edges_one_to_many` instance to class property methods

PR #193, opened 2023-11-20
updated cdf space name regex pattern

nikokaoja:


doctrino (approved):
I think we should reconsider if we should do these types of validations.

PR #209, opened 2023-11-23
Feature skip neat validators

nikokaoja:
## Added

- All neat specific validators for `Rules` can be now skipped by specifying them in `validators_to_skip`


CimStordal (approved):

cognite/neat/rules/models/rules.py
```diff
@@ -248,7 +254,12 @@ def convert_to_string(cls, value):
         return str(value)
 
     @validator("prefix", always=True)
-    def is_prefix_compliant(cls, value):
+    def is_prefix_compliant(cls, value, values):
+        if isinstance(values["validators_to_skip"], list) and (
+            inspect.currentframe().f_code.co_name in values["validators_to_skip"]
+            or "all" in values["validators_to_skip"]
```
doctrino, line 260: This is brutally hard to follow. Is there an easier design for this feature?

doctrino (commented):
I really like this feature, but when you throw in `inspect.currentframe().f_code.co_name` it will get very hard to maintain. Is there another way of implementing this that does not rely on very esoteric function calls to inspect?

PR #217, opened 2023-12-05
Feature from strings to things part 2

nikokaoja:
## Added
- Value types are now resolved as `ValueType` object instances


cognite/neat/app/api/routers/rules.py
```diff
@@ -90,9 +93,11 @@ def get_rules(
             }
             for value in rules.classes.values()
         ]
+
     except Exception as e:
         error_text = str(e)
 
+    print(properties)
```
doctrino, line 100: Debug?

doctrino (approved):
I did a quick skim of the code. Can go into more detail if needed, but this seems like primarily a refactoring.

PR #220, opened 2023-12-18
Feature from strings to things part 4

nikokaoja:
- Turning `ParentClass` string into `Entity`
- Added new fields to `Class` and `Property` as last step to enable advance data modeling
- Removed two validators from `Rules` which would otherwise block advance data modeling, specifically referring to `View`s and/or `Container`s in different spaces


cognite/neat/rules/exporter/_rules2dms.py
```diff
@@ -63,22 +63,29 @@ class DMSExporter(BaseExporter[DMSSchema]):
         container_policy: How to create/reuse existing containers.
         existing_model: In the case of updating an existing model, this is the existing model.
         report: Report. This is used when the exporter object is created from RawRules
+
+    !!! note "Container policy"
+        Here is more information about the different container policies:
+        - `create`: assumes no containers exist in CDF, will attempt to create them
+        - `reuse`: assumes containers exists in CDF, will attempt to only re-use them
+        - `extend`: will re-use existing, extend and/or create only missing containers
+        - `optimize`: create container of ideal size
```
doctrino, line 72: I would reformulate this, as idea size is very hard to claim. Can we say something else like `create containers trying to group based on Neat's optimization critera`.

cognite/neat/rules/models/_base.py
```diff
@@ -107,6 +107,18 @@ def from_list(cls, entity_strings: list[str], base_prefix: str | None = None, **
         ]
 
 
+class ParentClass(Entity):
+    type_: EntityTypes = EntityTypes.class_
```
doctrino, line 111: If I understand this correctly, then this should never be set on instantiation of the class, but be a fixed value for all instances. 
```suggestion
    type_: ClassVar[EntityTypes] = EntityTypes.class_
```
Remember to import `from typing import ClassVar`

cognite/neat/rules/models/_base.py
```diff
@@ -107,6 +107,18 @@ def from_list(cls, entity_strings: list[str], base_prefix: str | None = None, **
...
+
+
+class Container(Entity):
+    type_: EntityTypes = EntityTypes.container
+
+
+class ContainerProperty(Entity):
+    type_: EntityTypes = EntityTypes.property_
```
doctrino, line 119:
```suggestion
    type_: ClassVar[EntityTypes] = EntityTypes.container


class ContainerProperty(Entity):
    type_: ClassVar[EntityTypes] = EntityTypes.property_
```

cognite/neat/rules/models/rules.py
```diff
@@ -330,6 +310,29 @@ def is_suffix_compliant(cls, value, values):
...
             return value
 
+    @validator("namespace", always=True)
+    @skip_field_validator("validators_to_skip")
+    def set_namespace_if_none(cls, value, values):
+        if value is None:
+            if values["prefix"] == values["suffix"]:
+                return Namespace(f"http://purl.org/cognite/{values['prefix']}#")
+            else:
+                return Namespace(f"http://purl.org/cognite/{values['prefix']}/{values['suffix']}#")
```
doctrino, line 320: I find this slightly more readable, up to you if you agree. 
```suggestion
        if value is None:
            suffix = f"/{values['suffix']}" if values["prefix"] != values["suffix"] else ""
            return Namespace(f"http://purl.org/cognite/{values['prefix']}{suffix}#")

```

cognite/neat/rules/models/rules.py
```diff
@@ -330,6 +310,29 @@ def is_suffix_compliant(cls, value, values):
...
+        try:
+            return Namespace(parse_obj_as(HttpUrl, value))
```
doctrino, line 322: `parse_obj_as` is from `pydantic` `v1` lets use a `TypeAdapter`: https://docs.pydantic.dev/latest/migration/#introduction-of-typeadapter

cognite/neat/rules/models/rules.py
```diff
@@ -330,6 +310,29 @@ def is_suffix_compliant(cls, value, values):
         else:
             return value
 
+    @validator("namespace", always=True)
```
doctrino, line 313: In `pydantic v2`this is now `field_validatior` .

cognite/neat/rules/models/rules.py
```diff
@@ -553,29 +551,31 @@ def set_class_name_if_none(cls, value, values):
 
     @field_validator("parent_class", mode="before")
     @skip_field_validator("validators_to_skip")
-    def to_list_if_comma(cls, value, values):
+    def to_list_of_entities(cls, value, values):
         if isinstance(value, str):
             if value:
```
doctrino, line 556: No need to nest here
 ```suggestion
        if isinstance(value, str) and value: 
```

File: cognite/neat/rules/models/rules.py [hunk 8]
```diff
@@ -553,29 +551,31 @@ def set_class_name_if_none(cls, value, values):
...
+                    try:
+                        parent_classes.append(ParentClass.from_string(entity_string=v))
+                    # if all fails defaults "neat" object which ends up being updated to proper
+                    # prefix and version upon completion of Rules validation
+                    except ValueError:
+                        parent_classes.append(ParentClass(prefix="undefined", suffix=value, name=value))
+
+                return parent_classes
             if cls.model_fields[values.field_name].default is None:
                 return None
```
doctrino, line 568: This is now obsolete as if this `if` condition fails then if will also return None. 
```suggestion
```

cognite/neat/rules/models/rules.py
```diff
@@ -649,6 +649,12 @@ class Property(Resource):
     # Asset and Relationship at the same time
     cdf_resource_type: list[str] = Field(alias="Resource Type", default_factory=list)
 
+    # container specific things, only used for advance modeling or auto-filled by neat
+    container: Container | None = Field(alias="Container", default=None)
+    container_property: str | None = Field(alias="Container Property", default=None)
+    index: bool | None = Field(alias="Index", default=False)
+    constraints: str | None = Field(alias="Constraints", default=None, min_length=1)
```
doctrino, line 656: The reason why we need to have this here is that we need to have some way of expressing the physical storage of the data? Which thus have to be done on a model level.

cognite/neat/rules/models/rules.py
```diff
@@ -657,24 +663,27 @@ def is_raw_lookup(self) -> bool:
...
-    @validator("expected_value_type", pre=True)
-    def replace_string_with_entity(cls, value):
-        # handles simple types
+    @field_validator("container", mode="before")
+    def container_string_to_entity(cls, value):
+        if value:
+            try:
+                return Container.from_string(entity_string=value)
+            except ValueError:
+                return Container(prefix="undefined", suffix=value, name=value)
```
doctrino, line 672: My personal preference is to avoid indent as much as possible
```suggestion
        if not value:
            return value
        try:
            return Container.from_string(entity_string=value)
        except ValueError:
            return Container(prefix="undefined", suffix=value, name=value
```

doctrino (approved):
Mostly suggestions.

Issue #213, opened 2023-11-30
Moving from multi definition of Triple object to single resusable definition 


nikokaoja:
**Is your feature request related to a problem? Please describe.**
Currently there are **6** definitions of `Triple` object:

- `graph.stores.graph_store.Triple` as `TypeAlias = tuple[URIRef, URIRef, Literal | URIRef]`
- `graph.stores.oxrdflib._Triple` as `TypeAlias = tuple[Node, Node, Node]`
- ~~`graph.stores.transformations.Triple` as `dataclass` consisting of pars `subject`, `predicate` and `object`~~
- ~~`graph.loaders.rdf_to_dms.Triple` as `TypeAlias = tuple[Node, Node, Node]`~~
- ~~`rules.models.rdfpath.Triple` as `dataclass` consisting of pars `subject`, `predicate`, `object` and `optional`~~
- `utils.utils.Triple` as `TypeAlias = tuple[URIRef, URIRef, Literal | URIRef]`

**Describe the solution you'd like**
There should be only one definition, probably as pydantic model with set of methods that can turn model instance in desired form, such as `tuple` for consumption by `rdflib`

**Describe alternatives you've considered**
There is no alternative to clean and crisp code.

**Additional context**
This is grooming work.

nikokaoja:
Reduced from 6 -> 4, will go down to 2 (pydantic model + typealias)

PR #225, opened 2024-01-02
Graph Module Extractors Interface

doctrino:
I ended up moving `rdf_file_to_graph` to the stores module. I think it does not belong with the extractors as it is not extracting anything but simply loading an already serialized graph.


cognite/neat/graph/extractors/_base.py
```diff
@@ -0,0 +1,12 @@
...
+from collections.abc import Iterable
+
+
+class BaseExtractor:
+    """This is the base class for all extractors. It defines the interface that
+    extractors must implement.
+    """
+
+    @abstractmethod
+    def extract(self) -> Iterable[tuple]:
```
nikokaoja, line 11: Extract should always return `Iterable[tuple[URIRef, URIRef, URIRef|Literal]]`. Checkout issue:
https://github.com/cognitedata/neat/issues/213

cognite/neat/graph/extractors/_graph_capturing_sheet.py
```diff
@@ -59,7 +78,11 @@ def extract_triples_from_sheet(self, separator: str = ",", namespace: str | None
         """
         if self.filepath is None:
             raise ValueError("File path to the graph capturing sheet is not provided!")
-        return extract_graph_from_sheet(self.filepath, self.rules, separator, namespace)
+        graph_capturing_sheet = read_graph_excel_file_to_table_by_name(self.filepath)
```
nikokaoja, line 81: I am wandering whether we should have an abstract `read` or `read_source` (or some similar name) method that would read source `object` from which we extract triples?

nikokaoja (approved):
See my two comments, otherwise LGTM

cognite/neat/graph/extractors/_graph_capturing_sheet.py
doctrino: Lets keep the interface minimal for now until we see a strong need for it. You can always add methods to specific implementations.

PR #229, opened 2024-01-04
Fix for resetting graphdb graph using library step

msolvik:
Using the standard library step ResetGraphStores or the "reset" init_procedure of the ConfigureGraphStore step, only works for Oxigraph. 
This update adds functionality to reset GraphDB graphs as well.


cognite/neat/workflows/steps/lib/graph_store.py
```diff
@@ -252,4 +252,8 @@ def reset_store(
             graph_store.reinit_graph()
         elif data_store_dir:
             drop_graph_store_storage(store_type, data_store_dir, force=True)
+    elif store_type == RdfStoreType.GRAPHDB:
+        if graph_store:
+            graph_store.drop()
+            graph_store.reinit_graph()
```
doctrino, line 258: Can you update this to the new structure of store?

doctrino (changes requested):

docs/CHANGELOG.md
```diff
@@ -16,6 +16,11 @@ Changes are grouped as follows:
 - `Security` in case of vulnerabilities.
 
 
+## [0.54.0] - 04-01-24
+
+## Added
+- Reset graph option for GraphDBStore
+
 ## [0.TBD] - TBD-01-24
```
doctrino, line 24:
```suggestion
## [0.54.0] - 04-01-24

## Added
- Reset graph option for GraphDBStore

```

cognite/neat/graph/stores/_graphdb_store.py
```diff
@@ -47,5 +47,5 @@ def _set_graph(self) -> None:
 
     def drop(self):
         """Drops the graph."""
-        r = requests.delete(f"{self.graph_db_rest_url}/repositories/{self.graph_name}/rdf-graphs/service?default")
```
doctrino, line 50: This is the only place `graph_db_rest_usl` is used, can you double check and if it is, delete the attribute.

doctrino (approved):

cognite/neat/graph/stores/_graphdb_store.py
msolvik: What about here? 
https://github.com/cognitedata/neat/blob/main/cognite/neat/workflows/steps/lib/graph_store.py#L131

PR #237, opened 2024-01-16
Interface Loader Module

doctrino:
## Suggested interface
Introduce two interfaces to `loaders`
```python
class BaseLoader(ABC, Generic[T_Output]):
    """Base class for all loaders.

    A loader is a class that loads data from a source graph into
    target outside Neat.
    """

    def __init__(self, rules: Rules, graph_store: NeatGraphStoreBase):
        self.rules = rules
        self.graph_store = graph_store

    @overload
    def load(self, stop_on_exception: Literal[True]) -> Iterable[T_Output]:
        ...

    @overload
    def load(self, stop_on_exception: Literal[False] = False) -> Iterable[T_Output | ErrorDetails]:
        ...

    @abstractmethod
    def load(self, stop_on_exception: bool = False) -> Iterable[T_Output | ErrorDetails]:
        """Load the graph with data."""
        pass
```

Here `T_Output` is the data type the loader is loading.

In addition, the extended interface for loading to CDF.

```python
class CogniteLoader(BaseLoader[T_Output], ABC):
    """Base class for all loaders.

    A loader is a class that loads data from a source graph into
    target outside Neat.
    """

    @abstractmethod
    def load_to_cdf(
        self, client: CogniteClient, batch_size: int | None = 1000, max_retries: int = 1, retry_delay: int = 3
    ) -> None:
        """Load the graph with data."""
        pass
```

## One issue

In the original loaders the load and upload to cdf are setup in two separate steps. This is problematic as it requires all nodes/edges (for DMS) to be held in memory. This new implementation supports only keeping batch_size in memory at any point in time.


nikokaoja (approved):
This looks clean and very readable.

PR #240, opened 2024-01-20
AssetLoader

doctrino:
## Changes

* Implemented `AssetLoader` which is a reimplementation of `rdf2assets` to comply to the new loader interface.
* All configuration has been moved to the `__init__` method of `AssetLoader` meaning now you can, for example, turn off labels.
* Validated the new `AssetLoader` vs `rdf2assets` on a test case, in addition, manually gone through all the logic to ensure 
  all corner cases are covered.
* Removed usage of `pandas` and `pydantic` as it only obfuscates the logic. 
* Introduced a `.queries` property to `NeatBaseStore` to gather standard queries for easy reuse in one place. 
* **No changelog** entry. This is because the new `AssetLoader` is not to replace `rdf2assets` yet
    * It needs to support relationships as well.
    * Implement the `load_to_cdf` method.
    
Will save reminding implementation to a separate PR this can go in without changing the behavior of `neat`.


tests/app/api/test_workflows/graph_to_asset_hierarchy_workflow.yml
```diff
@@ -5347,13 +5347,13 @@ assets:
     IdentifiedObject.mRID: f17695aa-9aeb-11e5-91da-b8763fd99c5f
     IdentifiedObject.name: FORSMARK
     Substation.Region: f17695af-9aeb-11e5-91da-b8763fd99c5f
-    Substation.Terminal: 2dd901c6-bdfb-11e5-94fa-c8f73332c8f4, 2dd90258-bdfb-11e5-94fa-c8f73332c8f4,
```
doctrino, line 5350: I changed the original loader to sort metadata key values of list to ensure deterministic results. This is the result of that.

cognite/neat/graph/loaders/core/rdf_to_assets.py
```diff
@@ -174,6 +173,9 @@ def _define_asset_class_mapping(transformation_rules: Rules) -> dict[str, dict[s
                         asset_class_mapping[class_][resource_type_property] += [property_.property_name]
 
                     if property_.property_name not in asset_class_mapping[class_]["metadata"]:
+                        # Todo; Why Nikola?  This adds for example name property to metadata? Isn't that
+                        #   controlled by the resource_type_property? If you would like this behavior you
+                        #   would set resource_type_property to ["metadata", "name"]?
```
doctrino, line 178: This threw me for a loop. Do you know why this is done?

cognite/neat/graph/loaders/core/models.py
```diff
@@ -67,7 +67,7 @@ def preprocess_fields(cls, values: dict):
                 msg += f"has multiple values {values[field]}, "
                 msg += f"these values will be joined in a single string: {', '.join(values[field])}"
                 logging.info(msg)
-                values[field] = ", ".join(values[field])[: METADATA_VALUE_MAX_LENGTH - 1]
+                values[field] = ", ".join(sorted(values[field]))[: METADATA_VALUE_MAX_LENGTH - 1]
```
doctrino, line 70: Need deterministic behavior of the current loader to be able to compare. Since the query is not sorted these values can come in an arbitrary order.

tests/graph/loaders/test_asset_loader.py
```diff
@@ -0,0 +1,40 @@
...
+
+from cognite.neat.graph.loaders import AssetLoader
+from cognite.neat.graph.loaders.core.rdf_to_assets import rdf2assets
+from cognite.neat.graph.stores import MemoryStore
+from cognite.neat.rules.models import Rules
+
+
+class TestAssetLoader:
+    @pytest.mark.freeze_time("2024-01-01")
+    def test_vs_existing(self, transformation_rules: Rules, solution_knowledge_graph: Graph):
```
doctrino, line 14: Chech that the new `AssetLoader` does the same as the previous.

tests/graph/loaders/test_asset_loader.py
```diff
@@ -0,0 +1,40 @@
...
+        actual_dumped: dict[str, dict[str, Any]] = {}
+        for asset in loaded:
+            dumped = asset.dump(camel_case=False)
+            if asset.labels:
+                dumped["labels"] = [label.external_id for label in asset.labels]
+            if "description" not in dumped:
+                dumped["description"] = None
+            if "parent_external_id" not in dumped:
+                dumped["parent_external_id"] = None
+            actual_dumped[asset.external_id] = dumped
```
doctrino, line 31: Needed to reorg the output of the new one to match the output of `expected_assets`.

cognite/neat/graph/loaders/core/rdf_to_assets.py
nikokaoja: This was due to early requests by our customers that by default all properties end up in metadata. 
This was a quick fix (at that time) to make sure that all properties end up as keys in `Asset.metadata`

However, I think we should enforce explicit configuration, and start removing implicit behavior of neat.

nikokaoja: feel free to remove this implicit behavior, make sure that we state it in changelog

cognite/neat/graph/loaders/core/rdf_to_assets.py
```diff
@@ -378,7 +380,7 @@ def rdf2assets(
     # Step 4: Get ids of classes
     logging.info("Get ids of instances of classes")
     assets: dict[str, dict[str, Any]] = {}
-    class_ids = {class_: _get_class_instance_ids(graph, class_, namespace) for class_ in asset_class_mapping}
+    class_ids = {class_: _get_class_instance_ids(graph, namespace[class_]) for class_ in asset_class_mapping}
```
nikokaoja, line 383: smart !

cognite/neat/graph/stores/_base.py
```diff
@@ -317,3 +318,42 @@ def __iter__(self) -> Iterator[Triple]:
         prom_qsm.labels("query").observe(elapsed_time)
         prom_sq.labels("query").set(elapsed_time)
         return cast(Iterator[Triple], iter(result))
+
+
+class _Queries:
```
nikokaoja, line 323: This looks great! Nice to have them bundle up.

nikokaoja (approved):
Rdf2asset and rdf2relationships were really developed under pressure to provide a fast value to customers. This reflects in rather complicated code base, for which we never spent time to refactor.

This now looks clean and easy to maintain!

PR #241, opened 2024-01-21
Relationship loader

doctrino:


cognite/neat/graph/loaders/_asset_loader.py
```diff
@@ -126,16 +131,16 @@ def load(self, stop_on_exception: bool = False) -> Iterable[AssetResource | Erro
             properties_by_class_name[prop.class_id].append(prop)
 
         if self._use_labels:
+            self._loaded_labels.add(self._non_historic_label)
```
nikokaoja, line 134: Since none-historic / historic label is a very specific to a paricular customer, would it be more "generic" if we use for example of the following pairs:
Active - Inactive
Operational - Non-operational
Live - Dormant
Functioning - Deactivated
Online - Offline
Working - Idle
Engaged - Disengaged
Enabled - Disabled
Ongoing - Stopped
Operational - Inert
...

nikokaoja (approved):

cognite/neat/graph/loaders/_asset_loader.py
doctrino: Lets make this another PR.

PR #244, opened 2024-01-24
Aaido 159 add ability to dump data model to container.yaml and views.yaml in the step

nikokaoja:
## Added

- Added `ExportDMSDataModel` to dump data model (views) and containers as YAML

## Improved

- `DMSDataModelFromRules` is now extended such that one can update space/external_id/version of data model


cognite/neat/workflows/steps/lib/rules_exporter.py
```diff
@@ -49,9 +77,71 @@ def run(self, transformation_rules: RulesData) -> (FlowMessage, DMSDataModel):
...
+            version=data_model_contract.data_model.version,
+            description=data_model_contract.data_model.description,
+            name=data_model_contract.data_model.name,
+            views=list(data_model_contract.data_model.views.values()),
+        )
+
+        containers = dm.ContainerApplyList(data_model_contract.data_model.containers.values())
+
+        container_full_path.write_text(containers.dump_yaml())
+        data_model_full_path.write_text(data_model.dump_yaml())
```
doctrino, line 129: Is it a requirement that we use as few files as possible when it is dumped? 

When working with data models as YAML files my experience is that they quickly become very large and hard to maintain. So suggest we follow the standard from `cognite-toolkit` modules of one file per container, per view, and the data model is only specifying which view ids it expects. In addition, we use the prefix, `.container.yaml`, `.view.yaml`, and `data_model.yaml` to indicate what type of resource this is.

See for an example of what I suggest for output. https://github.com/cognitedata/cdf-project-templates/tree/main/cognite_toolkit/cognite_modules/examples/cdf_apm_simple_data_model/data_models

doctrino (commented):
I think neat should match `cognite-toolkit` expected input when we dump data models.

cognite/neat/workflows/steps/lib/rules_exporter.py
nikokaoja: Great! I will update it accordingly !

doctrino (approved):

PR #261, opened 2024-02-14
extended information rules capabilities

nikokaoja:
Added checks on:
- rule
- default
- source
- handful of @peoperty shortcuts that were previously been stored as fields


cognite/neat/rules/models/_rules/information_rules.py
```diff
@@ -164,28 +166,99 @@ class InformationProperty(DomainProperty):
...
+        """Type of property based on value type. Either data (attribute) or object (edge) property."""
+        if self.value_type.type_ == EntityTypes.data_value_type:
+            return EntityTypes.data_property
+        elif self.value_type.type_ == EntityTypes.object_value_type:
+            return EntityTypes.object_property
+        else:
+            return EntityTypes.undefined
+
+    @property
+    def isMandatory(self) -> bool:
```
doctrino, line 253:
```suggestion
    def is_mandatory(self) -> bool:
```

cognite/neat/rules/models/_rules/information_rules.py
```diff
@@ -164,28 +166,99 @@ class InformationProperty(DomainProperty):
...
+        """Returns True if property is mandatory."""
+        return self.min_count != 0
+
+    @property
+    def isList(self) -> bool:
```
doctrino, line 258:
```suggestion
    def is_list(self) -> bool:
```

tests/tests_units/rules/test_models/test_information_rules.py
```diff
@@ -46,6 +46,53 @@ def invalid_domain_rules_cases():
         "Value error, Metadata.role should be equal to 'information architect'",
         id="invalid_role",
     )
+    yield pytest.param(
+        {
+            "Metadata": {
+                "role": "information architect",
+                "creator": "Jon, Emma",
+                "contributor": "David",
```
doctrino, line 54: Is there a reason to distinguish between creator and contributor?

doctrino (approved):
Just prefer following Python convention of snake case :)

tests/tests_units/rules/test_models/test_information_rules.py
nikokaoja: Good point, we can simply call them all creator!

PR #262, opened 2024-02-16
Neat 21 round 3 information architect rules

nikokaoja:
First pass of moving validators to _type


cognite/neat/rules/models/_rules/_types.py
```diff
@@ -13,16 +57,126 @@
...
+            raise exceptions.ClassSheetParentClassIDRegexViolation(
+                illegal_ids, class_id_compliance_regex
+            ).to_pydantic_custom_error()
+    return value
+
+
+ParentClass_ = Annotated[
+    list[ParentClass] | None,
+    BeforeValidator(lambda v: split_parent(v)),
+    AfterValidator(lambda v: check_parent(v)),
```
doctrino, line 131: No-op lambda is no point :) 
```suggestion
    BeforeValidator(split_parent),
    AfterValidator(check_parent),
```

cognite/neat/rules/models/_rules/_types.py
```diff
@@ -13,16 +57,126 @@
...
+                # prefix and version upon completion of Rules validation
+                parents.append(ParentClass(prefix="undefined", suffix=v, name=v))
+
+        return parents
+    else:
+        return None
+
+
+def check_parent(value: list[ParentClass]) -> list[ParentClass]:
+    if value:
```
doctrino, line 114: Suggest early exit and avoid nested if
```suggestion
    if not value:
        return value
```

cognite/neat/rules/models/_rules/_types.py
```diff
@@ -13,16 +57,126 @@
...
+    custom_error(
+        lambda field_name, error, value: exceptions.VersionRegexViolation(
+            value, version_compliance_regex
+        ).to_pydantic_custom_error()
+    ),
+]
+
+
+def split_parent(value: str) -> list[ParentClass]:
+    if isinstance(value, str) and value:
```
doctrino, line 98:
```suggestion
    if not (isinstance(value, str) and value):
        return None
```

cognite/neat/rules/models/_rules/_types.py
```diff
@@ -13,16 +57,126 @@
...
+            if ENTITY_ID_REGEX_COMPILED.match(v) or VERSIONED_ENTITY_REGEX_COMPILED.match(v):
+                parents.append(ParentClass.from_string(entity_string=v))
+            else:
+                # if all fails defaults "neat" object which ends up being updated to proper
+                # prefix and version upon completion of Rules validation
+                parents.append(ParentClass(prefix="undefined", suffix=v, name=v))
+
+        return parents
+    else:
+        return None
```
doctrino, line 110:
```suggestion
    parents = []
    for v in value.replace(", ", ",").split(","):
        if ENTITY_ID_REGEX_COMPILED.match(v) or VERSIONED_ENTITY_REGEX_COMPILED.match(v):
             parents.append(ParentClass.from_string(entity_string=v))
        else:
            # if all fails defaults "neat" object which ends up being updated to proper
            # prefix and version upon completion of Rules validation
            parents.append(ParentClass(prefix="undefined", suffix=v, name=v))

    return parents
```

cognite/neat/rules/models/_rules/_types.py
```diff
@@ -13,16 +57,126 @@
...
+    StringConstraints(pattern=version_compliance_regex),
+    custom_error(
+        lambda field_name, error, value: exceptions.VersionRegexViolation(
+            value, version_compliance_regex
+        ).to_pydantic_custom_error()
+    ),
+]
+
+
+def split_parent(value: str) -> list[ParentClass]:
```
doctrino, line 97: What about making this private and put it at the end of the file to make it clear that types is what you should import from this module?
```suggestion
def _split_parent(value: str) -> list[ParentClass]:
```

cognite/neat/rules/models/_rules/_types.py
```diff
@@ -1,6 +1,50 @@
...
+from .base import (
+    class_id_compliance_regex,
+    more_than_one_none_alphanumerics_regex,
+    prefix_compliance_regex,
+    property_id_compliance_regex,
+    version_compliance_regex,
+)
+
+
+def custom_error(exc_factory: Callable[[str | None, Exception], Exception]) -> Any:
```
doctrino, line 35: Suggest making all functions private, (prefix underscore), and moving them to the end of the file to make it clear that you should imports the annotated types form this module.

cognite/neat/rules/models/_rules/domain_rules.py
```diff
@@ -32,93 +18,17 @@ class DomainMetadata(BaseMetadata):
...
-    class_: ExternalId = Field(alias="Class", min_length=1, max_length=255)
-    property_: ExternalId = Field(alias="Property", min_length=1, max_length=255)
-    value_type: ValueType = Field(alias="Value Type")
+    class_: Class_ = Field(alias="Class")
+    property_: Property_ = Field(alias="Property")
+    value_type: ValueType_ = Field(alias="Value Type")
     min_count: int | None = Field(alias="Min Count", default=None)
     max_count: int | float | None = Field(alias="Max Count", default=None)
 
-    @field_validator("value_type", mode="before")
```
doctrino, line 41: Beautiful :)

cognite/neat/rules/models/_rules/domain_rules.py
```diff
@@ -32,93 +18,17 @@ class DomainMetadata(BaseMetadata):
...
-                    "parent", ", ".join(illegal_ids)
-                ).to_pydantic_custom_error()
-            if illegal_ids := [v for v in value if not re.match(class_id_compliance_regex, v.suffix)]:
-                for v in illegal_ids:
-                    print(v.id)
-                raise exceptions.ClassSheetParentClassIDRegexViolation(
-                    illegal_ids, class_id_compliance_regex
-                ).to_pydantic_custom_error()
-        return value
+    parent: ParentClass_ = Field(alias="Parent Class")
```
doctrino, line 31: This does not collide with Python keyword, so no need to suffix and underscore. 
```suggestion
    parent: ParentClass = Field(alias="Parent Class")
```

cognite/neat/rules/models/_rules/information_rules.py
```diff
@@ -42,17 +33,10 @@ class MatchType(StrEnum):
...
-        description="This is used as prefix for generation of RDF OWL/SHACL data model representation",
-    )
-    namespace: Namespace = Field(
-        description="This is used as RDF namespace for generation of RDF OWL/SHACL data model representation "
-        "and/or for generation of RDF graphs.",
-        min_length=1,
-        max_length=2048,
-    )
+    prefix: Prefix
+    namespace: Namespace_
```
doctrino, line 39: Same here, no collision with Python keyword or built-in
```suggestion
    namespace: Namespace
```

cognite/neat/rules/models/_rules/information_rules.py
```diff
@@ -165,7 +123,11 @@ class InformationProperty(DomainProperty):
     """
 
     # TODO: Can we skip rule_type and simply try to parse the rule and if it fails, raise an error?
-
+    class_: Class_ = Field(alias="Class")
+    property_: Property_ = Field(alias="Property")
+    value_type: ValueType_ = Field(alias="Value Type")
```
doctrino, line 128: No collision with python build-in or keyword. 
```suggestion
    value_type: ValueType = Field(alias="Value Type")
```

doctrino (approved):
Very nice cleanup :) Just a few suggestions

cognite/neat/rules/models/_rules/domain_rules.py
nikokaoja: It collides with `ParentClass` , which is a class, while `ParentClass_` is a type we define.

cognite/neat/rules/models/_rules/information_rules.py
nikokaoja: Same here like in case of ParentClass

cognite/neat/rules/models/_rules/information_rules.py
nikokaoja: It colides since Namespace is rdflib entity, while Namespace_ is type we define. but I have renamed all the types in _type with suffix Type

cognite/neat/rules/models/_rules/_types.py
nikokaoja: Done

cognite/neat/rules/models/_rules/_types.py
nikokaoja: These are the same, there is no difference

cognite/neat/rules/models/_rules/_types.py
nikokaoja: I like the idea, but when moving them at the bottom it throws error:
`NameError: name '_split_parent' is not defined`

PR #272, opened 2024-02-24
[NEAT-28] Extend DMS rules to lists

doctrino:
Three changes to better cover all DMS possibilities:

1. The ValueType of the DMS rules sheet is now parsed as a `ViewEntity` if the relation column is set. This means we now support referring to views which are in a different space than the default set in the metadata sheet, and an arbitrary version.

The next are more esoteric, less important, but gives us a better cover of DMS rules
1. A property can now be used in multiple indexes. 
2. A property can now be part om multiple uniqueness constraints. 
3. A container can have a required constraint on multiple other containers.

### DIFF
--- a/cognite/neat/rules/models/_rules/dms_architect_rules.py
+++ b/cognite/neat/rules/models/_rules/dms_architect_rules.py
@@ -15,6 +15,7 @@ from cognite.neat.rules.models._rules.information_rules import InformationMetada
 
 from ._types import (
     ContainerEntity,
+    ContainerListType,
     ContainerType,
     ExternalIdType,
     PropertyType,
@@ -115,8 +116,8 @@ class DMSProperty(SheetEntity):
     class_: str = Field(alias="Class")
     property_: PropertyType = Field(alias="Property")
     description: str | None = Field(None, alias="Description")
-    value_type: str = Field(alias="Value Type")
-    relation: str | None = Field(None, alias="Relation")
+    value_type: ViewType | str = Field(alias="Value Type")
+    relation: Literal["direct", "multiedge"] | None = Field(None, alias="Relation")
     nullable: bool | None = Field(default=None, alias="Nullable")
     is_list: bool | None = Field(default=None, alias="IsList")
     default: str | int | dict | None | None = Field(None, alias="Default")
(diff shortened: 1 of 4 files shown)


cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -125,46 +127,60 @@ class DMSProperty(SheetEntity):
     container_property: str | None = Field(None, alias="ContainerProperty")
     view: ViewType | None = Field(None, alias="View")
     view_property: str | None = Field(None, alias="ViewProperty")
-    index: str | None = Field(None, alias="Index")
-    constraint: str | None = Field(None, alias="Constraint")
+    index: StrListType | None = Field(None, alias="Index")
+    constraint: StrListType | None = Field(None, alias="Constraint")
+
+    @field_validator("value_type", mode="before")
```
nikokaoja, line 133: I guess this could not fit under `_types.py` ?

cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -125,46 +127,60 @@ class DMSProperty(SheetEntity):
...
+        if not isinstance(value, str):
+            return value
+
+        if info.data.get("relation"):
+            # If the property is a relation (direct or edge), the value type should be a ViewEntity
+            # for the target view (aka the object in a triple)
+            return ViewEntity.from_raw(value)
+        return value
+
+    @field_serializer("value_type", when_used="unless-none")
```
nikokaoja, line 144: same as above

cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -125,46 +127,60 @@ class DMSProperty(SheetEntity):
...
 
 
 class DMSContainer(SheetEntity):
     class_: str | None = Field(None, alias="Class")
     container: ContainerType = Field(alias="Container")
     description: str | None = Field(None, alias="Description")
-    constraint: ContainerType | None = Field(None, alias="Constraint")
+    constraint: ContainerListType | None = Field(None, alias="Constraint")
 
     def as_container(self, default_space: str) -> dm.ContainerApply:
```
nikokaoja, line 157: nice

nikokaoja (approved):

cognite/neat/rules/models/_rules/dms_architect_rules.py
doctrino: It is only used here, which is why I do not put it into the types.

cognite/neat/rules/models/_rules/dms_architect_rules.py
doctrino: Same argument

PR #296, opened 2024-03-07
GraphStore optimisations and other improvments

alivinco:
## Fixed
- Fixed issue with prefixes not being updated during GraphStore (oxi) reinitialisation
- Fixed graph store reset issue for JSON loader
- Small UI adjustments

## Added
- Added rules browser to the UI. Now user can browse all rules in UI from local store .
- Added configurable HTTP headers for `DownloadDataFromRestApiToFile` step. The feature is useful when the API requires specific headers to be set (has been requested by Cognite customer).


cognite/neat/workflows/steps/lib/graph_store.py
```diff
@@ -237,21 +239,17 @@ def run(  # type: ignore[override]
             reset_store(store_dir, graph_store.graph if graph_store else None)
             logging.info("Graph reset complete")
 
+        prefixes = rules_data.rules.prefixes if rules_data else PREFIXES
```
doctrino, line 242: Lets avoid mutating by accident the constant
```suggestion
        prefixes = rules_data.rules.prefixes if rules_data else PREFIXES.copy()
```

cognite/neat/app/api/routers/rules.py
```diff
@@ -28,17 +34,17 @@ def get_rules(
...
+        workflow = NEAT_APP.workflow_manager.get_workflow(workflow_name)
+        if workflow is None:
+            return {"error": f"Workflow {workflow_name} is not found"}
+        workflow_definition = workflow.get_workflow_definition()
+        if not file_name:
+            for step in workflow_definition.steps:
+                if step.method == "ImportExcelToRules":
+                    file_name = step.configs["file_name"]
+                    version = step.configs["version"]
+                    break
```
doctrino, line 47: I see this is just moving code that is already existing, but this seems like a hack to have special treatment of a step.

cognite/neat/app/api/routers/rules.py
```diff
@@ -28,17 +34,17 @@ def get_rules(
...
-                version = step.configs["version"]
-                break
+    if workflow_name != "undefined":
+        workflow = NEAT_APP.workflow_manager.get_workflow(workflow_name)
+        if workflow is None:
+            return {"error": f"Workflow {workflow_name} is not found"}
+        workflow_definition = workflow.get_workflow_definition()
+        if not file_name:
+            for step in workflow_definition.steps:
+                if step.method == "ImportExcelToRules":
```
doctrino, line 44: I would import the `ImportExcelToRules`, just to make sure that the dependency is traceable in code. 
```suggestion
                if step.method == ImportExcelToRules.__name__:
```

doctrino (commented):
A few minor code suggestions, more comments coming wrt to the UI changes

doctrino (approved):

PR #297, opened 2024-03-10
[NEAT-60, NEAT-61] Added Zip importer and Support v2 Spec

doctrino:
Features
* Support DTDL spec v2 in `DTDLImporter`
* Added `from_zip` to the `DTDLImporter`
* In `DTDLImporter`, if an object is missing in the loaded spec for a relationship. A warning is triggered and the target value 
   type is a JSON instead of pointing to the unknown object.
* Extended `DTDLImporter` to handle `Commands` and `Telemetry` objects.
* Added validation on `DMSRules` on the required containers existing if the schema is complete.

Fixes
* In `DMSExporter`, if a view has changed, delete and recreate the data model to trigger the GraphQL conversion.
* Fix bug in `DMSExporter` reporting created and changed when these failed.
* Bug in `as_schema` of `DMSRules` using default space instead of view spaces for edge type.
* Fix `DMSExporter` reporting container has changed if `usedFor` is not set to default.


nikokaoja (approved):

cognite/neat/rules/importers/_dtdl2rules/_unit_lookup.py
```diff
@@ -0,0 +1,224 @@
...
+@dataclass
+class UnitEntry:
+    semantic_type: str
+    unit_type: str
+    unit: str
+
+
+_UNIT_ENTRIES = [
+    UnitEntry("Acceleration", "AccelerationUnit", "centimetrePerSecondSquared"),
+    UnitEntry("Acceleration", "AccelerationUnit", "gForce"),
```
nikokaoja, line 13: How much is this related to QUDT, which our units in CDF are based on?
It is fine to have as is here, but in general I would be so happy to see that CDF comes pre-builded with controlled vocab of units which are interconnected with different standards:
1. QUDT (as base)
2. DTDL

...

cognite/neat/rules/validation/__init__.py
```diff
@@ -30,8 +30,9 @@
     MissingView,
 )
```
nikokaoja, line 31: Rename `Validation` to `Issues`

nikokaoja (approved):

cognite/neat/rules/validation/__init__.py
doctrino: Will do in later PR

PR #298, opened 2024-03-11
fix: interface of Base/CDF Exporter

doctrino:
Two fixes
* CDFExporter is a type of BaseExporter
* It looks like the example `config.yaml` have a real secret. Make it clear that this is fake

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -1,6 +1,7 @@
 cdf_client:
     project: get-power-grid
     client_id: ca4b8f9c6-7d6e-4d7b-9cf3-6d4a32f8b7e1
+    # This is a fake secret, replace with your own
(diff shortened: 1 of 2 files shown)


docs/config.yaml
```diff
@@ -1,7 +1,8 @@
 cdf_client:
     project: get-power-grid
     client_id: ca4b8f9c6-7d6e-4d7b-9cf3-6d4a32f8b7e1
-    client_secret: Zy7!nM4cFp9sH3gR6tB8kL0oP7eU6wD5vQ4zN9yF
```
doctrino, line 4: This look like a cause for calling an incident, even though it it a mock secret.

nikokaoja (approved):

PR #312, opened 2024-03-15
[NEAT-31] Conversion workflow

doctrino:
* Add `output_role` parameter `Excelporter` such that the user can chose a different output than the input format.
* Added this as as a new configuration to the step.
* Extended the `ValidationRules` with a conversion step. 

image

### DIFF
--- a/cognite/neat/rules/exporters/_base.py
+++ b/cognite/neat/rules/exporters/_base.py
@@ -6,6 +6,7 @@ from typing import Generic, TypeVar
 from cognite.client import CogniteClient
 
 from cognite.neat.rules._shared import Rules
+from cognite.neat.rules.models._rules import DMSRules, InformationRules, RoleTypes
 
 from ._models import UploadResult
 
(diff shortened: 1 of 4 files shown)


cognite/neat/rules/exporters/_base.py
```diff
@@ -21,6 +22,16 @@ def export_to_file(self, filepath: Path, rules: Rules) -> None:
     def export(self, rules: Rules) -> T_Export:
         raise NotImplementedError
 
+    def _convert_to_output_role(self, rules: Rules, output_role: RoleTypes | None = None) -> Rules:
```
nikokaoja, line 25: NICE !!!!

cognite/neat/workflows/steps/lib/rules_exporter.py
```diff
@@ -145,36 +146,50 @@ class RulesToExcel(Step):
             label="Styling of the Excel file",
             options=list(exporters.ExcelExporter.style_options),
         ),
+        Configurable(
+            name="Output role format",
```
nikokaoja, line 150: great that you are also using userfriendly names for UI

nikokaoja (approved):

PR #313, opened 2024-03-15
Neat 83 workflow visualization of information rules using mock data. part 3 workflow updated docs

nikokaoja:
- Added GraphFromMockGraph step
- Added  Visualize Data Model using Mock graph workflow
- Small update when exporting DMS rules to InfoArchitect

Screenshot 2024-03-15 at 13 17 36


doctrino (commented):
I struggle to generate more than the default one node of each type. 

I try to change the class counts, but it is still just one node generated for each class:
image

doctrino (approved):
Approving, but think we probably need a nice way to upload rules for how to generate mock data.

PR #314, opened 2024-03-15
[NEAT-62] Cleanup Validation -> Issues Module

doctrino:
Most of the code is moving round existing classes and renaming. That accounts for most of the changes in code-lines.

Changes:

* Rename `rules.valiadion` to `rules.issues`
* Remove all imports in `__init__.py` and make all submodules of the new `rules.issues` public such that you can use the modules for navigation. E.g. `from cognite.neat.rules import issues` and then `issues.dms.MissingSpacesError`.
* All Errors have the suffix `Error` and all warnings have suffix `Warning`. Introduced test to enforce this standard.
* Introduce a base class for all error and warning types, this is used when formatting the error.


cognite/neat/rules/issues/__init__.py
```diff
@@ -0,0 +1,27 @@
+from . import dms, fileread, formatters, importing, spreadsheet, spreadsheet_file
```
nikokaoja, line 1: Nice org

cognite/neat/rules/issues/spreadsheet.py
```diff
@@ -0,0 +1,381 @@
...
+        )
+
+    def dump(self) -> dict[str, Any]:
+        output = super().dump()
+        output["view_id"] = self.view_id
+        return output
+
+
+@dataclass(frozen=True, order=True)
+class ClassNoPropertiesNoParentsWarning(ValidationWarning):
```
nikokaoja, line 239: There are couple of issues which depending on the use case can be warning or error. Is this possible with the current setup?

nikokaoja (approved):

cognite/neat/rules/issues/spreadsheet.py
doctrino: Then, you would have to duplicate have one error and one warning version of the issue.

PR #316, opened 2024-03-16
[NEAT-108] Transformation/RawTable/RawDatabase Loaders

doctrino:


cognite/neat/utils/cdf_loaders/__init__.py
```diff
@@ -0,0 +1,25 @@
+from ._base import ResourceLoader
```
nikokaoja, line 1: Should this at some point migrate under a special type of loaders under rules?

cognite/neat/utils/cdf_loaders/_base.py
```diff
@@ -0,0 +1,62 @@
...
+)
+from cognite.client.data_classes.data_modeling.ids import (
+    InstanceId,
+    VersionedDataModelingId,
+)
+from cognite.client.utils.useful_types import SequenceNotStr
+
+from .data_classes import RawTableID
+
+T_ID = TypeVar("T_ID", bound=str | int | DataModelingId | InstanceId | VersionedDataModelingId | RawTableID)
```
nikokaoja, line 23: What about some more readable name .e.g. `T_CogniteResourceId`?

cognite/neat/utils/cdf_loaders/_base.py
```diff
@@ -0,0 +1,62 @@
...
+    Generic[T_ID, T_WriteClass, T_WritableCogniteResource, T_CogniteResourceList, T_WritableCogniteResourceList],
+):
+    resource_name: str
+
+    def __init__(self, client: CogniteClient) -> None:
+        self.client = client
+
+    @classmethod
+    @abstractmethod
+    def get_id(cls, item: T_WriteClass | T_WritableCogniteResource) -> T_ID:
```
nikokaoja, line 38: suggestion maybe rename this to `as_id` ? wandering do you need two separate methods one for listable, one for single items? can there be a single `as_id` method?

nikokaoja: even better call this `as_external_ids` to comply with sdk

cognite/neat/utils/cdf_loaders/data_classes.py
```diff
@@ -0,0 +1,118 @@
...
+from typing import Any, cast
+
+from cognite.client import CogniteClient
+from cognite.client.data_classes._base import (
+    CogniteResourceList,
+    WriteableCogniteResource,
+    WriteableCogniteResourceList,
+)
+
+# The Table, TableWrite data classes in the Cognite-SDK lacks the database attribute.
```
nikokaoja, line 12: Would this eventually find its way to `cognite-sdk` ?

nikokaoja (approved):

cognite/neat/utils/cdf_loaders/_base.py
doctrino: Great, will include.

cognite/neat/utils/cdf_loaders/_base.py
doctrino: Hmm, I want to distinguish this from `.as_id()`, `.as_ids()` which are on most Cognite Resource objects.

cognite/neat/utils/cdf_loaders/data_classes.py
doctrino: Yes, hopefully, but the problem is that would be a breaking change (to the poor RAW implementation), so it would be breaking the SDK.

PR #318, opened 2024-03-16
[NEAT-110] Support direct relation list

doctrino:

PR #317, opened 2024-03-16
[NEAT-108] Export transformations Part 2

doctrino:
image

image

image

PR #319, opened 2024-03-17
[NEAT-74] In Model Option 

doctrino:
Merge after #315 , #316 , #317 , #318 

Adds a boolean column `InModel` to the view sheet. This enables the user to have views in their sheet, typically base views that are used in implements of other views, that are not included as part of the data model.


cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -204,6 +204,7 @@ def from_container(cls, container: dm.ContainerApply) -> "DMSContainer":
 class DMSView(SheetEntity):
     view: ViewType = Field(alias="View")
     implements: ViewListType | None = Field(None, alias="Implements")
+    in_model: bool = Field(True, alias="InModel")
```
nikokaoja, line 207: nice, this is what we discussed about stating certain views to be created but not added to the data model

nikokaoja (approved):

PR #321, opened 2024-03-17
[Prep for NEAT-48] Refactor: _DMSExporter

doctrino:
This is refactoring the `_DMSExporter` by gathering and moving out the methods for creating the containers, views, and spaces. 

This is necessary as with adding filter to the views the code became very messy. Dragging this out as its own PR to separate the refactoring from actually adding the feature.


cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -496,89 +496,66 @@ def __init__(
...
+        views: dm.ViewApplyList,
+        data_model: dm.DataModelApply,
+    ) -> dm.SpaceApplyList:
+        used_spaces = {container.space for container in containers} | {view.space for view in views}
+        if len(used_spaces) == 1:
+            # We skip the default space and only use this space for the data model
+            data_model.space = used_spaces.pop()
+            spaces = dm.SpaceApplyList([dm.SpaceApply(space=data_model.space)])
+        else:
+            spaces = dm.SpaceApplyList([metadata.as_space()] + [dm.SpaceApply(space=space) for space in used_spaces])
```
nikokaoja, line 544: doesn't DMSMetadata have field `space` already, why calling method?

cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -652,38 +625,109 @@ def to_schema(self, rules: DMSRules) -> DMSSchema:
...
+                continue
+            for prop in container_properties:
+                if prop.container_property is None:
+                    continue
+                if isinstance(prop.value_type, DMSValueType):
+                    type_cls = prop.value_type.dms
+                else:
+                    type_cls = dm.DirectRelation
+
+                prop_name = to_camel(prop.container_property) if self.standardize_casing else prop.container_property
```
nikokaoja, line 680: For time being this is fine, but we already have under SheetEntity defined option to add `name` field, in case for example when property_ and class_ (i.e. ids) are human unreadable, which is the case for one of our clients.

nikokaoja (approved):

cognite/neat/rules/models/_rules/dms_architect_rules.py
doctrino: To get the space object

cognite/neat/rules/models/_rules/dms_architect_rules.py
doctrino: Not sure if I understand this. Are you suggesting a better way to handle properties?

PR #322, opened 2024-03-17
[NEAT-43] Add reverse direct relations

doctrino:
Merge after #315 #316 #317 #318 #319 #320 #321

For reverse direct relations, I need to know which property it is the reverse of. To solve this problem I introduce `ViewPropEntity`  class. This is a `ViewEntity` that also holds a property. I switched the `value_type` in the `DMSProperty` to use this instead.


cognite/neat/rules/models/_rules/_types/_field.py
```diff
@@ -247,12 +248,14 @@ def _semantic_value_type_before_validator(value: Any) -> Any:
 ]
 
 CdfValueType = Annotated[
-    DMSValueType | ViewEntity,
+    DMSValueType | ViewPropEntity,
```
nikokaoja, line 251: is this correct? ValueType can be either another View or simple values, I am not sure how it can be Value Property?

cognite/neat/rules/models/_rules/_types/_base.py
```diff
@@ -222,6 +227,60 @@ def as_id(
         return ViewId(space=space, external_id=external_id, version=version)
 
 
+class ViewPropEntity(ViewEntity):
```
nikokaoja, line 230: if this was made to figure out which properties are "reversed" , maybe reusing methods from `_analysis` would be useful:
- get_class_linkage, to get class linkage
- get_symmetric_pairs got get which classes are linked with symmetric/reverse properties

cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -165,6 +166,26 @@ def direct_relation_must_be_nullable(cls, value: Any, info: ValidationInfo) -> N
...
+            raise ValueError(f"Reverse direct relation must not have container or container property, got {value}")
+        return value
+
+    @field_validator("value_type", mode="after")
+    def relations_value_type(cls, value: CdfValueType, info: ValidationInfo) -> CdfValueType:
+        if (relation := info.data["relation"]) is None:
+            return value
+        if not isinstance(value, ViewPropEntity):
+            raise ValueError(f"Relations must have a value type that points to another view, got {value}")
+        if relation == "reversedirect" and value.property_ is None:
```
nikokaoja, line 181: I see you are storing here information about reverse property, which should be called `symmetric property`

cognite/neat/rules/models/_rules/_types/_base.py
```diff
@@ -222,6 +227,60 @@ def as_id(
...
+    type_: ClassVar[EntityTypes] = EntityTypes.view_prop
+    property_: str | None = None
```
nikokaoja, line 232: A bit ugly way to store symmetric property, we can fix this later. 

But rename `property_` to `symmetric_property` since If I am not mistaken you are storing reverse property here, not actual property that points to this View.

tests/tests_unit/rules/test_models/test_dms_architect_rules.py
```diff
@@ -403,6 +403,220 @@ def rules_schema_tests_cases() -> Iterable[ParameterSet]:
...
+                    value_type="text",
+                    container="Asset",
+                    container_property="name",
+                    view="Asset",
+                    view_property="name",
+                ),
+                DMSProperty(
+                    class_="Asset",
+                    property_="timeseries",
+                    value_type="Timeseries:asset",
```
nikokaoja, line 432: Here we have View `Asset` that has property `timeseries` which points to ? `Timeseries` (simple type), but you want to add reverse property `asset` to `Timeseries` as object to write back? 

My worry here is that it breaks the pattern `space:View(version)` when referring to views. suddenly we have BaseType:reverseProperty as pattern, so my proposal would be to have following:

`BaseType(reverseProperty=...)` and `space:View(version=..., reverseProperty=...)

cognite/neat/rules/models/_rules/dms_architect_rules.py
nikokaoja: or simply call it reversDirect or reverseProperty

nikokaoja (commented):
Main issue is with breaking the pattern due to a need to have a reverseProperty in the regex. 

My worry here is that it breaks the pattern space:View(version) when referring to views. suddenly we have `BaseType:reverseDirect` as pattern, so my proposal would be to have following:

BaseType(reverseProperty=...) and `space:View(version=..., reverseProperty=...)

cognite/neat/rules/models/_rules/_types/_field.py
doctrino: See PR description.

tests/tests_unit/rules/test_models/test_dms_architect_rules.py
doctrino: Lets not include `reverse` here as that is already part of the relation. Anyways I am all ears for good suggestions, I think this is maybe easier to reuse `space:View(version=..., property=...)`.

nikokaoja (approved):

PR #323, opened 2024-03-17
Debug loading Core Spec

doctrino:
Fixed bugs while debugging the core spec
* Ensure excel file is released by `ExcelImporter`
* Ensure the `ExcelImporter` capture warnings correctly.
* Errors raised by pydantic is handled appropriately by the `ExcelImporter` (before it only worked for errors inheriting from `InvalidSheetError`
* Added helpers `errors` and `warnings` to `IssueList` makes it easy to see warning and errors in a notebook.
* All strings to the rules get stripped. (A very hard error to figure out why the regex is failing due to one extra space at the end of `Describable `.
* If a container had no properties, but another required it, I kept the container. Discovered that this is not possible as the CDF API do not allow empty containers, so instead removed the required constraint on empty containers. (Empty containers happens as we convert list of direct relations to edges to overcome that this is missing in the API)


nikokaoja (approved):
Nice!

PR #326, opened 2024-03-18
[NEAT-115] Add support for node type

doctrino:
Adds `Filter` column that can be either `hasData`, `nodeType`, or `hasData,nodeType`. 

Starting intentionally with minimum of filters supported, lets instead wait for people to request more before we add it.

### DIFF
--- a/cognite/neat/rules/issues/dms.py
+++ b/cognite/neat/rules/issues/dms.py
@@ -339,7 +339,7 @@ class UnsupportedFilterWarning(DMSSchemaWarning):
     filter_type: str
 
     def message(self) -> str:
-        return f"The filter {self.filter_type} in {self.view_id} is not supported. This filter will be ignored."
+        return f"The filter {self.filter_type} in {self.view_id} is not supported. This filter will be ignored. Contact the neat team for more information on custom filters."
 
     def dump(self) -> dict[str, Any]:
         output = super().dump()
--- a/cognite/neat/rules/issues/dms.py
+++ b/cognite/neat/rules/issues/dms.py
@@ -339,7 +339,10 @@ class UnsupportedFilterWarning(DMSSchemaWarning):
     filter_type: str
 
     def message(self) -> str:
-        return f"The filter {self.filter_type} in {self.view_id} is not supported. This filter will be ignored. Contact the neat team for more information on custom filters."
+        return (
+            f"The filter {self.filter_type} in {self.view_id} is not supported. This filter will be ignored. "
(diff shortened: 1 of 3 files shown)


cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -701,29 +709,39 @@ def _create_views_with_node_types(
...
-            else:
-                # Child filter with its own container properties
-                if node_type_flag:
-                    # Transformations do not yet support setting node type.
-                    view.filter = dm.filters.And(has_data, node_type)
-                    node_types.append(dm.NodeApply(space=view.space, external_id=view.external_id, sources=[]))
+            selected_filters = dms_view_by_view_id[view.as_id()].filter_
+
+            filters: list[dm.filters.Filter] = []
+            for selected_filter in selected_filters or ["hasData"]:
```
nikokaoja, line 722: Probably misreading the code but this at first reading looks like you can add either "hasData" or "nodeType" but not both.

cognite/neat/rules/issues/dms.py
```diff
@@ -310,3 +310,39 @@ def dump(self) -> dict[str, Any]:
...
+
+@dataclass(frozen=True)
+class UnsupportedFilterWarning(DMSSchemaWarning):
+    description = "The filter type is not supported by neat"
+    fix = "Change the filter to a supported type"
+    view_id: dm.ViewId
+    filter_type: str
+
+    def message(self) -> str:
+        return f"The filter {self.filter_type} in {self.view_id} is not supported. This filter will be ignored."
```
nikokaoja, line 342:
```suggestion
        return f"The filter {self.filter_type} in {self.view_id} is not supported. This filter will be ignored. Contact the neat team for more information on custom filters."
```

nikokaoja (approved):

cognite/neat/rules/models/_rules/dms_architect_rules.py
doctrino: It is clearly a list, which is appended to, so not really sure how to make it much clearer that this is a set of filters.

doctrino, line 722:
```suggestion
            default_filters = ["hasData"]
            for selected_filter in selected_filters or default_filters:
```

doctrino: Did a small trick now that hopefully makes this clearer.

PR #329, opened 2024-03-19
improve base pydantic errors in issues

nikokaoja:
Small update (one liner) that bubbles up what failed. Previously it was only saying "Field required" since `loc` was not considered in the error
Screenshot 2024-03-19 at 15 28 25


doctrino (approved):
Elegant, just 6 lines of code

PR #345, opened 2024-03-25
Neat 125 dms rules information rules. data types which are not found are always treated as a reference even for misspelling such as sting 

nikokaoja:
### Improved
- Input for InformationArchitect XSD value types are now case insensitive.


Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: run-explorer run-tests run-linters build-ui build-python build-docker run-docker compose-up
 
-version="0.68.5"
+version="0.69.0"
```
doctrino, line 3: Maybe wait with a release until after the DMS is also fixed? Is this urgent?

And if this is a release, shouldn't it be `0.68.6`?

cognite/neat/rules/models/_rules/_types/_field.py
```diff
@@ -259,9 +261,9 @@ def _semantic_value_type_serializer(value: Any) -> str:
 CdfValueType = Annotated[
     DMSValueType | ViewPropEntity,
     BeforeValidator(
-        lambda value: DMS_VALUE_TYPE_MAPPINGS[value]
-        if value in DMS_VALUE_TYPE_MAPPINGS
-        else ViewPropEntity.from_raw(value)
+        lambda value: (
+            DMS_VALUE_TYPE_MAPPINGS[value] if value in DMS_VALUE_TYPE_MAPPINGS else ViewPropEntity.from_raw(value)
```
doctrino, line 265: Shouldn't there be a lower here?
```suggestion
            DMS_VALUE_TYPE_MAPPINGS[value] if value.lower() in DMS_VALUE_TYPE_MAPPINGS else ViewPropEntity.from_raw(value)
```

doctrino (approved):
Just a few questions

Makefile
nikokaoja: I agree, let's hold version until DMS is fixed.

PR #348, opened 2024-03-25
added test bump version

nikokaoja:
### Improved
- Input for DmsArchitect DMS value types are now case insensitive.


cognite/neat/rules/importers/_dms2rules.py
```diff
@@ -92,6 +92,8 @@ def to_rules(
                             direct_value_type = "UNKNOWN"
                         else:
                             direct_value_type = ViewPropEntity.from_id(prop.source)
+
+                        print(direct_value_type, type(direct_value_type))
```
doctrino, line 96: Debug?
```suggestion
```

doctrino (approved):

cognite/neat/rules/importers/_dms2rules.py
nikokaoja: Hm I do not have that locally interesting

PR #356, opened 2024-04-01
Neat 138 Anders Review

doctrino:
The main change is that I use a class based approach instead of a functional one. The advantage is that you avoid having to pass around a lot of arguments. In addition, we avoid introducing another Enum that we need to do multiple if checks on.


cognite/neat/rules/importers/_spreadsheet2rules.py
```diff
@@ -77,43 +171,38 @@ def to_rules(
...
-
-        rules = cast(DomainRules | InformationRules | DMSRules, user_rules or reference_rules)
+            rules = user_rules
+            rules.reference = reference_rules
+        elif user_rules:
+            rules = user_rules
+        elif reference_rules:
+            rules = reference_rules
+        else:
+            raise ValueError("No rules found.")
```
doctrino, line 205: I added this part, as I think we want to pass on the reference rules with the user rules if they are both present. However, this implementation does not trigger validation, so we might have to rethink this a bit. But waiting to get feedback on the suggestion in this PR.

PR #358, opened 2024-04-03
Move sql types

doctrino:
Realized that the part using the SQL types, so added a test to ensure that the code was hit.

### DIFF
--- a/cognite/neat/rules/models/_rules/_types/_value.py
+++ b/cognite/neat/rules/models/_rules/_types/_value.py
@@ -29,6 +29,7 @@ class ValueTypeMapping(BaseModel):
     python: type
     dms: type
     graphql: str
+    sql: str
 
 
 # mypy: ignore-errors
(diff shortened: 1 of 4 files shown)


cognite/neat/rules/models/_rules/_types/_value.py
```diff
@@ -67,42 +73,54 @@ def __str__(self) -> str:
...
+        "GraphQL": "TimeSeries",
+        "dms": TimeSeriesReference,
+        "SQL": "STRING",
+    },
+    {"name": "file", "python": FileReference, "GraphQL": "File", "dms": FileReference, "SQL": "STRING"},
+    {"name": "sequence", "python": SequenceReference, "GraphQL": "Sequence", "dms": SequenceReference, "SQL": "STRING"},
+    {"name": "json", "python": Json, "GraphQL": "Json", "dms": Json, "SQL": "STRING"},
 ]
 
 _DMS_TYPES: list[dict] = [
```
doctrino, line 106: I am a bit confused why there are two constants here

nikokaoja: prob. fastest way to make DMS complaint value types.

nikokaoja (approved):

PR #366, opened 2024-04-06
[NEAT-156] Line endings

doctrino:
**Problem**: By default, when `Python` write to file it uses the default line endings of the OS system. This leads to problem when using neat to generate data models which are checked into Git history, as two users with different OSes, will get a massive diff due to the line endings.

**Suggested solution**: Neat is opinionated about the line endings (and encodings) to ensure consistent user experience between different OSes.

**Why hide it?** On the importers this is a hidden configuration (private class attribute). The reason for this is that importers are made for the most basic user that should not try to change this. However, in an edge case when someone wants to change it, they can although with no guarantee from **NEAT** that it will continue to work :)

### DIFF
--- a/cognite/neat/rules/exporters/_base.py
+++ b/cognite/neat/rules/exporters/_base.py
@@ -14,6 +14,9 @@ T_Export = TypeVar("T_Export")
 
 
 class BaseExporter(ABC, Generic[T_Export]):
+    _new_line = "\n"
+    _encoding = "utf-8"
+
     @abstractmethod
     def export_to_file(self, rules: Rules, filepath: Path) -> None:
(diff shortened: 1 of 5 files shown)


cognite/neat/rules/exporters/_base.py
```diff
@@ -14,6 +14,9 @@
 
 
 class BaseExporter(ABC, Generic[T_Export]):
+    _new_line = "\n"
+    _encoding = "utf-8"
```
nikokaoja, line 18: Famously infamous line-ending challenge.
Great!

nikokaoja (approved):

PR #372, opened 2024-04-06
[NEAT-147] Information to DMS with Reference

doctrino:
Implemented conversion of Information to DMS: 
 * Property reference -> Implemented as a container references
 * View references -> Used to implements. 
 
and added DMS sheet for Olav.


cognite/neat/rules/models/_rules/information_rules.py
```diff
@@ -465,29 +463,24 @@ def _as_dms_property(cls, prop: InformationProperty) -> "DMSProperty":
...
+            property_=prop.property_,
+            value_type=value_type,
+            nullable=nullable,
+            is_list=is_list,
+            relation=relation,
+            default=prop.default,
+            reference=prop.reference,
+            container=container,
+            container_property=container_property,
+            view=ViewPropEntity(prefix=prop.class_.prefix, suffix=prop.class_.suffix, version=prop.class_.version),
```
nikokaoja, line 481: We assume here that Class id -> View id == Container id, which is something we need to highlight to end users, that in general modeling concepts there are no views or containers, but classes.

cognite/neat/rules/models/_rules/information_rules.py
```diff
@@ -499,3 +492,27 @@ def _to_space(cls, prefix: str) -> str:
...
+            return (
+                ContainerEntity(prefix=prop.reference.prefix, suffix=prop.reference.suffix),
+                prop.reference.property_ or prop.property_,
+            )
+        else:
+            return ContainerEntity(prefix=prop.class_.prefix, suffix=prop.class_.suffix), prop.property_
+
+    @classmethod
+    def _get_view_implements(cls, cls_: InformationClass) -> list[ViewEntity]:
+        if isinstance(cls_.reference, ReferenceEntity):
```
nikokaoja, line 508: They can use either parent class or reference, not both. Important to stress them out this in tutorial, because we need to assume that ppl will start mixing things up.

nikokaoja (approved):

PR #382, opened 2024-04-09
[NEAT-163] Unknown object.

doctrino:


cognite/neat/rules/issues/dms.py
```diff
@@ -183,7 +183,7 @@ def dump(self) -> dict[str, Any]:
 
 
 @dataclass(frozen=True)
-class DirectRelationMissingSourceError(DMSSchemaError):
+class DirectRelationMissingSourceWarning(DMSSchemaWarning):
```
doctrino, line 186: Treating this as an error becomes to strict as it will break import of any model in CDF with source not set in direct relation.

nikokaoja (approved):

PR #393, opened 2024-04-15
[NEAT-177] Reference edge type

doctrino:
**Problem** When referencing the edge.type should be set to the original type. This is not happening.

Here is the manual fixe needed in the customer project when trying to build a solution model and reuse a multiedge property, 
and being surprised when no edges showed up. 

image

### DIFF
--- a/cognite/neat/rules/models/_rules/dms_architect_rules.py
+++ b/cognite/neat/rules/models/_rules/dms_architect_rules.py
@@ -735,11 +735,22 @@ class _DMSExporter:
                             "Multiedge relation must have a view as value type. "
                             "This should have been validated in the rules"
                         )
-                    view_property = dm.MultiEdgeConnectionApply(
-                        type=dm.DirectRelationReference(
+                    if isinstance(prop.reference, ReferenceEntity):
+                        ref_view_prop = prop.reference.as_prop_id(
+                            default_space, default_version, self.standardize_casing
+                        )
+                        edge_type = dm.DirectRelationReference(
+                            space=ref_view_prop.source.space,
+                            external_id=f"{ref_view_prop.source.external_id}.{ref_view_prop.property}",
+                        )
+                    else:
+                        edge_type = dm.DirectRelationReference(
                             space=source.space,
                             external_id=f"{prop.view.external_id}.{prop.view_property}",
-                        ),
+                        )
+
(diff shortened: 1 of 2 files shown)


cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -735,11 +735,22 @@ def _create_views_with_node_types(
                             "Multiedge relation must have a view as value type. "
                             "This should have been validated in the rules"
                         )
-                    view_property = dm.MultiEdgeConnectionApply(
-                        type=dm.DirectRelationReference(
+                    if isinstance(prop.reference, ReferenceEntity):
+                        ref_view_prop = prop.reference.as_prop_id(
+                            default_space, default_version, self.standardize_casing
```
nikokaoja, line 740: I understood from the previous PR that `standardized_casing` was causing issue and that you removed it?

cognite/neat/rules/models/_rules/dms_architect_rules.py
```diff
@@ -804,9 +825,17 @@ def _create_views_with_node_types(
...
         for view in views:
             ref_containers = sorted(view.referenced_containers(), key=lambda c: c.as_tuple())
-            has_data = dm.filters.HasData(containers=list(ref_containers)) if ref_containers else None
-            node_type = dm.filters.Equals(["node", "type"], {"space": view.space, "externalId": view.external_id})
             dms_view = dms_view_by_id.get(view.as_id())
+            has_data = dm.filters.HasData(containers=list(ref_containers)) if ref_containers else None
+            if dms_view and isinstance(dms_view.reference, ReferenceEntity):
+                # If the view is a reference, we implement the reference view,
+                # and need the filter to match the reference
+                ref_view = dms_view.reference.as_id(False, default_space, default_version, self.standardize_casing)
```
nikokaoja, line 833: aren't we able to pick up space and version from reference ? would this lead to different space/version?

nikokaoja (approved):

cognite/neat/rules/models/_rules/dms_architect_rules.py
doctrino: I don't think this can happen in practice, but the method requires it. There is some technical debt around the entity classes.

PR #395, opened 2024-04-15
[NEAT-179] Remove standardize casing

doctrino:
Why was this introduced? I was lazy and did not want to fix it manually for a data model.

Why remove this?
  * Magically fixing casing confuses the end user.
  * It is cause for increased code complexity and opportunity for bugs (There has already been on bug). I think it is likely to happen again when you work with a reference model.
  * It should likely not be **Neat**'s task to be opinionated about casing in the data model.


cognite/neat/rules/exporters/_rules2dms.py
```diff
@@ -40,8 +40,6 @@ class DMSExporter(CDFExporter[DMSSchema]):
             If set, only export components in the given spaces. Defaults to None which means all spaces.
         existing_handling (Literal["fail", "skip", "update", "force"], optional): How to handle existing components.
             Defaults to "update". See below for details.
-        standardize_casing(bool, optional): Whether to standardize the casing. This means PascalCase for external ID
```
nikokaoja, line 43: Agree that this should not be `NEAT` feature, we can suggest in best practices to use `PascalCase` for classes and `camelCase` for properties but that is all

nikokaoja (approved):

cognite/neat/rules/exporters/_rules2dms.py
doctrino: Yes absolutely.

PR #397, opened 2024-04-16
[NEAT-182] Automatically create metadata sheet for reference model.

doctrino:
# Automatically set metadata sheet 

Adds one parameter `new_model_id` to the `ExcelImporter`. This is used to automatically fill out the previously empty `metadata` sheet when exporting a reference model.

**In addition** Set the metadata correctly when importing a data model from CDF. 

image


cognite/neat/rules/exporters/_rules2excel.py
```diff
@@ -194,3 +211,74 @@ def _adjust_column_widths(cls, workbook: Workbook) -> None:
...
+            metadata["prefix" if is_info else "space"] = self.new_model_id[0]  # type: ignore[index]
+            metadata["title" if is_info else "externalId"] = self.new_model_id[1]  # type: ignore[index]
+            metadata["version"] = self.new_model_id[2]  # type: ignore[index]
+        else:
+            metadata["prefix" if is_info else "space"] = existing_model_id[0]
+            metadata["title" if is_info else "externalId"] = existing_model_id[1]
+            metadata["version"] = existing_model_id[2]
+
+        if is_solution and is_info:
+            metadata["namespace"] = f"http://purl.org/{self.new_model_id[0]}/{self.new_model_id[1]}"  # type: ignore[index]
```
nikokaoja, line 252: do not add title ! as title can be any set of strings, just use prefix, ie:
f"http://purl.org/{self.new_model_id[0]}/"

nikokaoja (approved):

PR #398, opened 2024-04-16
Small feat hackathon

nikokaoja:
### Fixed
- `ExcelImporter` was resetting `role` value to value set in rules instead of keeping value provided as arg
### Changed
- Default namespace set to `http://purl.org/cognite/neat#` instead of `http://purl.org/cognite/app#`
- OwlImporter for rules v2 has `make_compliant` set to False by default
### Added
- When creating OWL from rules, prefix will be saved under property `neat:prefix` (hence change of default namespace)
- When reading OWL if there is `neat:prefix` value will be added to `rules.metadata.prefix`
- By default we are defaulting OWL properties to min 0 and max 1 occurrence if no occurrence is set
- Added test for generation of complete rules out of partial rules


cognite/neat/rules/importers/_spreadsheet2rules.py
```diff
@@ -171,7 +171,7 @@ def to_rules(
         is_reference: bool = False,
     ) -> tuple[Rules | None, IssueList] | Rules:
         issue_list = IssueList(title=f"'{self.filepath.name}'")
-
+        print(role)
```
doctrino, line 174: Debug

doctrino (approved):

PR #418, opened 2024-04-27
[NEAT-183] Refactor entities

doctrino:
## Why Refactor Entities.

* There is a lot of repeated code in the current Entities. 
* This implementation split parsing from validation. This enables us to have Warnings instead of Errors, ref NEAT-59, NEAT-79.
* This implementation has the logic in the base `Entity` class, with relatively easy extensions.

### DIFF
--- a/cognite/neat/rules/models/entities.py
+++ b/cognite/neat/rules/models/entities.py
@@ -196,6 +196,9 @@ class ClassEntity(Entity):
 class ParentClassEntity(ClassEntity):
     type_: ClassVar[EntityTypes] = EntityTypes.parent_class
 
+    def as_class_entity(self) -> ClassEntity:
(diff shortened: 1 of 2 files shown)


cognite/neat/rules/models/entities.py
```diff
@@ -0,0 +1,299 @@
+import re
```
nikokaoja, line 1: Is this renamed from `_entity.py` from `models` ?

nikokaoja (approved):

cognite/neat/rules/models/entities.py
doctrino: No, I made a new parallell to avoid breaking existing, then I will spend the next PR moving over to these new entities.

PR #420, opened 2024-04-28
[NEAT-183] Apply new entities

doctrino:
Note that 920 of the lines is just the test updates to new format `test_dms_architect_rules.py`

* Applied all new entities and data types. Deleted the old.
* The only left use of the old Entiry class is in rdfpath, so I moved that in there to keep the rdfpath contained.
* All data modeling entities, `ViewEntity`, `ContainerEntity`, ..., have space, external id and suffix required. The rational is that these are implementation and should thus always be a concrete. This have consequences for the `DMSRules` class, as it becomes very hard to instatiate.
* To overcome the above, I introduced a `DMSRulesWrite` dataclass which is used for input. The rational is that `DMSRules` before had two responsibilities, input data and validate. Now there is a separate class for inputting data.


tests/tests_unit/rules/test_analysis.py
```diff
@@ -71,4 +71,15 @@ def test_inherited_referred_classes(self, olav_rules: InformationRules) -> None:
...
+            ClassEntity(prefix="power", suffix="Point"),
+            ClassEntity(prefix="power", suffix="Polygon"),
+        }
+
+        actual = InformationArchitectRulesAnalysis(olav_rules).referred_classes
+
+        missing = expected - actual
+        assert not missing, f"Missing classes: {missing}"
+        extra = actual - expected
+        assert not extra, f"Extra classes: {extra}"
```
doctrino, line 85: I changed this test. It used to be 4 due to the fact that `ParentClassEntity` and `ClassEntity` used to be considered not the same entity:
image

Now, since `ParentClassEntity` is inheriting from `ClassEntity` they are considered equal, so `ParentClassEntity(prefix="power", suffix="GeoLocation")` is removed.

cognite/neat/rules/importers/_owl2rules/_owl2rules.py
```diff
@@ -126,15 +126,15 @@ def _add_missing_value_types(components: dict) -> dict:
...
+    xsd_types = _XSD_TYPES
     candidate_value_types = {definition["Value Type"] for definition in components["Properties"]} - {
         definition["Class"] for definition in components["Classes"]
     }
 
     # to avoid issue of case sensitivity for xsd types
     value_types_lower = {v.lower() for v in candidate_value_types}
 
-    xsd_types_lower = {x.lower() for x in xsd_types}
+    xsd_types_lower = {x.lower().removeprefix("xsd:") for x in xsd_types}
```
nikokaoja, line 137: this is nice method `removeprefix` was not aware it exists!

cognite/neat/rules/exporters/_rules2ontology.py
```diff
@@ -114,7 +115,7 @@ def from_rules(cls, input_rules: Rules) -> Self:
             ],
             shapes=[
                 SHACLNodeShape.from_rules(
-                    class_dict[class_.suffix],
+                    class_dict[str(class_.suffix)],
```
nikokaoja, line 118: Do you need to have `str(...)` ?

cognite/neat/rules/exporters/_rules2ontology.py
```diff
@@ -249,15 +250,15 @@ def from_class(cls, definition: InformationClass, namespace: Namespace, prefixes
             sub_class_of = []
             for parent_class in definition.parent:
                 try:
-                    sub_class_of.append(prefixes[parent_class.prefix][parent_class.suffix])
+                    sub_class_of.append(prefixes[str(parent_class.prefix)][str(parent_class.suffix)])
```
nikokaoja, line 253: same here

cognite/neat/rules/models/data_types.py
```diff
@@ -175,17 +190,17 @@ class DateTime(DataType):
...
 
 
-class DateTimeStamp(DataType):
-    name = "dateTimeStamp"
+class Timestamp(DataType):
+    name = "timestamp"
     python = datetime
     dms = dms.Timestamp
     graphql = "Timestamp"
     xsd = "xsd:dateTimeStamp"
```
nikokaoja, line 198: Def remove "xsd:" from all the types, since we know when .xsd that that is the namespace.
Alt we need almost everywhere to remove prefix "xsd:" from string.

cognite/neat/rules/exporters/_rules2ontology.py
```diff
@@ -324,12 +325,14 @@ def from_list_of_properties(cls, definitions: list[InformationProperty], namespa
...
             owl_property.type_.add(OWL[definition.type_])
-            owl_property.range_.add(
-                XSD[definition.value_type.suffix]
-                if definition.value_type.suffix in XSD_VALUE_TYPE_MAPPINGS
-                else namespace[definition.value_type.suffix]
-            )
-            owl_property.domain.add(namespace[definition.class_.suffix])
+
+            if isinstance(definition.value_type, DataType):
+                owl_property.range_.add(XSD[definition.value_type.xsd])
```
nikokaoja, line 330: This is causing issue since for example if we have `String` data_type, corresponding `.xsd` attribute returns `xsd:strign` which then when passed to `XSD` namespace creates URI `http://www.w3.org/2001/XMLSchema#xsd:string` which is ill formed URI.  Instead this method is suppose to generate `http://www.w3.org/2001/XMLSchema#string`

We have two options, either remove `xsd:` all toghether from all `data_type` definitions, or do XSD[definition.value_type.xsd.removeprefix("xsd:"). I think we should go with the first one (as just discussed in the call).

cognite/neat/rules/exporters/_rules2ontology.py
```diff
@@ -552,8 +555,8 @@ def from_property(cls, definition: InformationProperty, namespace: Namespace) ->
             node_kind=SHACL.IRI if definition.type_ == EntityTypes.object_property else SHACL.Literal,
             expected_value_type=(
                 namespace[f"{definition.value_type.suffix}Shape"]
-                if definition.type_ == EntityTypes.object_property
-                else XSD[definition.value_type.suffix]
+                if isinstance(definition.value_type, ClassEntity)
+                else XSD[definition.value_type.xsd]
```
nikokaoja, line 559: Same as above, `XSD[definition.value_type.xsd]` will produce illformed URI

cognite/neat/rules/models/data_types.py
```diff
@@ -207,7 +222,7 @@ class PlainLiteral(DataType):
     python = str
     dms = dms.Text
     graphql = "String"
-    xsd = "xsd:string"
+    xsd = "xsd:plainLiteral"
```
nikokaoja, line 225: same here, my bad for not suggesting this earlier when you introduced improved data_type handling!

nikokaoja (approved):

PR #421, opened 2024-04-29
Neat 206 delete dms data model step is missing in v2 rules

nikokaoja:
### Added
- `DMSExporter` now supports deletion of data model and data model components
- `DeleteDataModelFromCDF` added to the step library


cognite/neat/utils/cdf_loaders/_data_modeling.py
```diff
@@ -87,8 +87,10 @@ def retrieve(self, ids: SequenceNotStr[str]) -> SpaceList:
     def update(self, items: Sequence[SpaceApply]) -> SpaceList:
         return self.create(items)
 
-    def delete(self, ids: SequenceNotStr[str]) -> list[str]:
-        return self.client.data_modeling.spaces.delete(ids)
+    def delete(self, ids: SequenceNotStr[str] | Sequence[Space | SpaceApply]) -> list[str]:
```
doctrino, line 90: Suggest adding a technical debt task (unless you want to tackle it immediately) on moving this to the base class.

doctrino (approved):
This is good. As we discussed yesterday, the support for item and not just id should be in the base class of the loader, but that can be later.

If you have time, an integration test would be great for this.

PR #424, opened 2024-05-02
🪨Robustify 

doctrino:
Attempt at enabling parallelization of test execution. 

* Moved `Config` up from `app/api/data_classes/config` to `neat/config.py` as it is used in Workflows as well as app.
* Removed all hardcoded use of directories used by the `neat` app and workflows. This is much more robust and thus the reason I think this is worth merging.   

There are still 7 tests which does not work with parallel execution, but stopping this effort for now, to focus on more urgent features/fixes.


cognite/neat/config.py
```diff
@@ -1,39 +1,169 @@
...
+    from typing import Self
+else:
+    from backports.strenum import StrEnum
+    from typing_extensions import Self
+
+LOG_FORMAT = "%(asctime)s.%(msecs)03d %(levelname)-8s %(message)s"
+LOG_DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
+
+
+class RulesStoreType(StrEnum):
```
nikokaoja, line 27: NICE!

nikokaoja (approved):

PR #426, opened 2024-05-03
Neat 141 legacy rules UI switch

nikokaoja:
### Improved
- Steps are now categorized as `current`, `legacy`, and `io` steps
- Workflow fails if one mix `current` and `legacy` steps in the same workflow


doctrino (approved):
Need better UI error messages.

PR #433, opened 2024-05-05
[NEAT-222] DMS (and all other rules) dump order 🖶

doctrino:
Instead of writing any fancy logic in the individual exporters (for example `ExcelExporter`). 

I moved the properties `class_`, `name` and `description` from the base class `SheetEntity` to each derived class. This is a little bit of repeated properties, but it makes controlling the order that the fields are dumped much much easier. This is also more robust, as the current implementation dependend on `class` always being the first property.


cognite/neat/rules/models/rules/_dms_architect_rules.py
```diff
@@ -201,10 +204,10 @@ class DMSProperty(SheetEntity):
     reference: URLEntity | ReferenceEntity | None = Field(default=None, alias="Reference", union_mode="left_to_right")
```
nikokaoja, line 204: I am tempted to move `reference` after `constraint`

nikokaoja (approved):
This will remove so much issues from end users

PR #439, opened 2024-05-06
Neat 231 update docs to reflect default view version being dropped

nikokaoja:


doctrino (approved):
Except for the temporary excel files `~....xlsx` all good :)

PR #440, opened 2024-05-07
[NEAT-226] Introduce last 👋

doctrino:
A tiny PR to add the new field `Last` to ensure we are aligned on what we are planning to do.

Splitting the refactoring work if using this new field into multiple PRs.

### DIFF
--- a/cognite/neat/rules/models/dms/_converter.py
+++ b/cognite/neat/rules/models/dms/_converter.py
@@ -107,7 +107,8 @@ class _DMSRulesConverter:
             metadata=metadata,
             properties=SheetList[InformationProperty](data=properties),
             classes=SheetList[InformationClass](data=classes),
-            reference=self.dms.reference and self.dms.reference.as_information_architect_rules(),  # type: ignore[arg-type]
+            last=self.dms.last.as_information_architect_rules() if self.dms.last else None,
+            reference=self.dms.reference.as_information_architect_rules() if self.dms.reference else None,
         )
 
     @classmethod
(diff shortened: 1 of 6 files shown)


cognite/neat/rules/models/dms/_converter.py
```diff
@@ -107,7 +107,8 @@ def as_information_architect_rules(
             metadata=metadata,
             properties=SheetList[InformationProperty](data=properties),
             classes=SheetList[InformationClass](data=classes),
-            reference=self.dms.reference and self.dms.reference.as_information_architect_rules(),  # type: ignore[arg-type]
+            last=self.dms.last.as_information_architect_rules() if self.dms.last else None,
```
nikokaoja, line 110: I think we made a right choice of having added last, as we now have stack of three models. I imagine we can build rather strong UI lineage showing interconnection of these three.

cognite/neat/rules/models/dms/_rules.py
```diff
@@ -322,15 +323,6 @@ def matching_version_and_space(cls, value: SheetList[DMSView], info: ValidationI
             warnings.warn(issues.dms.ViewModelSpaceNotMatchingWarning(different_space, metadata.space), stacklevel=2)
         return value
 
-    @field_validator("views")
```
nikokaoja, line 325: Why is this removed?

nikokaoja (approved):

cognite/neat/rules/models/dms/_rules.py
doctrino: Bad merged, it is included in the validation above.

PR #446, opened 2024-05-10
Anders review

doctrino:
* Moving validation to the creation of the filter.
* Introducing a `constants.py` for a place to keep all hardcoded constants coming from prod wrt to DMS. Can probably be named better and have it another place.


doctrino:
Missing that DMS filters can be imported.

PR #449, opened 2024-05-11
[prep NEAT-236] Refactor DMS schema single model

doctrino:


cognite/neat/rules/importers/_dms2rules.py
```diff
@@ -179,21 +180,12 @@ def to_rules(
 
     def _create_rule_components(
         self,
+        data_model: dm.DataModelApply,
         schema: DMSSchema,
```
nikokaoja, line 184: would not schema have data model ? why it is necessary to additionally data_model as arg?

cognite/neat/rules/models/dms/_exporter.py
```diff
@@ -73,7 +73,7 @@ def to_schema(self) -> DMSSchema:
 
         output = DMSSchema(
             spaces=spaces,
-            data_models=dm.DataModelApplyList([data_model]),
+            data_model=data_model,
```
nikokaoja, line 76: this if not mistaken reverts back to the original idea of only having single data model

nikokaoja (approved):

PR #452, opened 2024-05-14
[NEAT-250] 🐛DMSImporter missing referenced views

doctrino:
It turns out that a data model in CDF can have views that points through other views that are not in the data model (and you might not have access to)


cognite/neat/rules/models/dms/_schema.py
```diff
@@ -146,8 +147,20 @@ def from_data_model(
         space_write = space_read.as_write()
 
         view_loader = ViewLoader(client)
-        # We need to include parent views in the schema to make sure that the schema is valid.
+
         existing_view_ids = set(views.as_ids())
+
+        # We need to include all views the edges/direct relations are pointing to have a complete schema.
```
nikokaoja, line 153: This is occurs if data model is not made by neat?

cognite/neat/rules/models/dms/_schema.py
```diff
@@ -203,6 +216,21 @@ def from_data_model(
             reference=ref_schema,
         )
 
+    @classmethod
+    def _connection_references(cls, view: dm.View) -> set[dm.ViewId]:
```
nikokaoja, line 220: We should raise warning that our data model is not complete since it is pointing to types outside of the data model.

nikokaoja (approved):

cognite/neat/rules/models/dms/_schema.py
doctrino: Yes, it could. In this case, it was actually created by a buggy version of Neat.

PR #453, opened 2024-05-14
[NEAT-240] 🏃Setting up metadata ExcelExporter

doctrino:
A few notes
* Moved out the conversion of Metadata sheets between information and dms for easy reuse.
* Discovered that there were several parameters in the `.as_information_architect_rules` and `.as_dms_architect_rules`. So removed them.

Overall some refactoring in this PR, the actual contribution is +60 lines, `_MetadataCreator`, and  +100 lines of tests


cognite/neat/rules/models/dms/_converter.py
```diff
@@ -32,31 +34,16 @@ def as_domain_rules(self) -> "DomainRules":
 
     def as_information_architect_rules(
         self,
-        created: datetime | None = None,
-        updated: datetime | None = None,
-        name: str | None = None,
-        namespace: Namespace | None = None,
```
doctrino, line 38: These were never set or used, so removed them as they are only causing confusion.

cognite/neat/rules/models/information/_converter.py
```diff
@@ -30,30 +30,18 @@ def __init__(self, information: InformationRules):
     def as_domain_rules(self) -> DomainRules:
         raise NotImplementedError("DomainRules not implemented yet")
 
-    def as_dms_architect_rules(self, created: datetime | None = None, updated: datetime | None = None) -> "DMSRules":
```
doctrino, line 33: Same issues here:
These were never set or used, so removed them as they are only causing confusion.

pyproject.toml
```diff
@@ -129,7 +129,7 @@ twine = "*"
 pytest = { version = ">=8.0, <8.1" }
 pytest-cov = "*"
 pytest-notebook = "*"
-pytest-freezegun = "*"
```
doctrino, line 132: This version is no longer maintained so switched it.

cognite/neat/rules/exporters/_rules2excel.py
```diff
@@ -234,82 +238,67 @@ def _adjust_column_widths(cls, workbook: Workbook) -> None:
...
-
-        if is_solution:
-            metadata["dataModelType"] = DataModelType.solution.value
+    def __init__(
+        self,
+        action: Literal["create", "update"],
+        new_model_id: tuple[str, str] | None = None,
+    ):
+        self.action = action
+        self.new_model_id = new_model_id or ("YOUR_PREFIX", "YOUR_TITLE")
```
nikokaoja, line 251: so in case of InformationRules prefix goes to space, title as an external id. 
My worry is that title can have characters not compliant with external_id of model.

nikokaoja: What makes new id for information model is actually `namespace`, as it is a unique id of data model.
So in case of Information model new id would be "YOUR_PREFIX" "YOUR_NAMESPACE"

nikokaoja: As it is set up now, maybe we are fine, as you are using `prefix` to create `namespace`

cognite/neat/rules/exporters/_rules2excel.py
```diff
@@ -234,82 +238,67 @@ def _adjust_column_widths(cls, workbook: Workbook) -> None:
...
+        return created
+
+    def _create_new_info(self, now: datetime) -> InformationMetadata:
+        prefix = self.new_model_id[0]
+        return InformationMetadata(
+            data_model_type=DataModelType.solution,
+            schema_=SchemaCompleteness.extended,
+            extension=ExtensionCategory.addition,
+            prefix=prefix,
+            namespace=f"http://purl.org/{prefix}/",  # type: ignore[arg-type]
```
nikokaoja, line 297:
```suggestion
            namespace=f"http://purl.org/{prefix}/",  # type: ignore[arg-type]
```
```suggestion
            namespace=f"http://purl.org/neat/{prefix}/",  # type: ignore[arg-type]
```

nikokaoja: As we own http://purl.org/neat/ , we own any subsub domains

cognite/neat/rules/models/information/_converter.py
```diff
@@ -111,6 +99,26 @@ def as_dms_architect_rules(self, created: datetime | None = None, updated: datet
...
+        )
+
+        space = cls._to_space(metadata.prefix)
+
+        return DMSMetadata(
+            schema_=metadata.schema_,
+            space=space,
+            data_model_type=metadata.data_model_type,
+            version=metadata.version,
+            external_id=metadata.name.replace(" ", "_").lower(),
```
nikokaoja, line 115: this works as long as someone does not use other characters in the name such as ":" , etc.

Maybe it is safe to have one more regex replacement for any non-complinat characters

nikokaoja (approved):

PR #459, opened 2024-05-19
[NEAT-237, NEAT-262] 🫠Adding Svein Harald Spreadsheets

doctrino:
Discovered that these sheets were missing in addition to a set of bugs.

Loading Svein Harald's sheet turned out to completely break the validation of DMS, so had to do NEAT-237 as well.

* Before this PR, when you dump with last you do not get the Metadata sheet. I have changed this to include it, as I think it shows the linage of the model, and it is also clearer in the validation which is performed on the last model.

Realize that there are several inconsistencies in the code wrt to the metadata properties. Suggest the following:

DataModelType this determines how we validate against the reference sheets.
1. **enterprise**. No validation is performed against the reference model, it can be there but is assumed to be external to CDF and cannot be validated against 
2. **solution**. There must be a Reference model.

SchemaCompleteness This determines the overall consistency checks for the Data Model. This is connected to the Last sheets.
1. **complete** Last sheets are ignored. All classes/views/containers should be in the user sheets.
2. **extended** Last sheet must be present, and it assumed that user+last = complete. With the additional validation determined by the extension field.
3. **partial**:  No consistency checks are performed on the model as a whole.

ExtensionType: 
1. **Rebuild**: Everything is allowed
2. **Reshape**: Containers cannot change.
3.  **Additition**: Containers and views cannot change'

**Note** New solution model should set their schema to `complete`. The schema as a whole will still include the reference model.


nikokaoja (approved):

cognite/neat/rules/models/dms/_validation.py
```diff
@@ -26,12 +27,17 @@ def __init__(self, rules: DMSRules):
         self.issue_list = IssueList()
 
     def validate(self) -> IssueList:
-        self._validate_best_practices()
+        self._validate_raw_filter()
         self._consistent_container_properties()
+
         self._referenced_views_and_containers_are_existing()
```
nikokaoja, line 33: Should we do any check if SchemaCompleteness is `partial` ?

nikokaoja (approved):

cognite/neat/rules/models/dms/_validation.py
nikokaoja: I think this method will break if .last is missing?

nikokaoja (approved):

cognite/neat/rules/models/dms/_validation.py
doctrino: I have this:
```
        if not self.rules.last:
            raise ValueError("The schema is set to 'extended', but no last rules are provided to validate against")
```

cognite/neat/rules/models/dms/_validation.py
```diff
@@ -26,12 +27,17 @@ def __init__(self, rules: DMSRules):
...
-        self._validate_extension()
-        self._validate_schema()
-        self._validate_performance()
+        if self.metadata.schema_ is SchemaCompleteness.extended:
+            self._validate_extension()
+        if self.metadata.schema_ is SchemaCompleteness.partial:
+            return self.issue_list
+        dms_schema = self.rules.as_schema()
+        self.issue_list.extend(dms_schema.validate())
```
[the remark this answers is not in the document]
doctrino: Yes, since it turned into a one liner, I thought it was overkill to have a function for it.

PR #472, opened 2024-05-30
Neat 269 importer data model inference step

nikokaoja:
### Added
- Added `RulesInferenceFromRdfFile` to the step library


doctrino (commented):
Same question as the last importer. Is the `make_compliant` a property of the importer? Seems relevant for all data models you import into neat.

PR #473, opened 2024-05-30
[NEAT-95] 🚛Draft DMS Loader

doctrino:
This is a first draft with the core skeleton of the DMS Loader

Moving to separate PR:
* Error handling
* Tracking of ingestion
* Writing to CDF
* Implement the `.as_read` method in the `DMSSchema`.

In addition, we already have a task on upgrading the store https://cognitedata.atlassian.net/browse/NEAT-279

### DIFF
--- a/cognite/neat/graph/loaders/_rdf2dms.py
+++ b/cognite/neat/graph/loaders/_rdf2dms.py
@@ -1,11 +1,15 @@
 import json
 import warnings
+from collections import defaultdict
 from collections.abc import Iterable
 from pathlib import Path
+from typing import cast, overload
 
 import yaml
 from cognite.client import CogniteClient
 from cognite.client import data_modeling as dm
+from pydantic import HttpUrl, TypeAdapter, ValidationError
+from rdflib.term import URIRef
 
 from cognite.neat.graph.stores import NeatGraphStoreBase
 from cognite.neat.rules.issues import NeatValidationError
@@ -16,10 +20,10 @@ from ._base import CDFLoader
 
 class DMSLoader(CDFLoader[dm.InstanceApply]):
     def __init__(
-        self, data_model: dm.DataModel[dm.View], graph_store: NeatGraphStoreBase, add_class_prefix: bool = False
+        self, graph_store: NeatGraphStoreBase, data_model: dm.DataModel[dm.View], add_class_prefix: bool = False
     ):
+        super().__init__(graph_store)
         self.data_model = data_model
-        self.graph_store = graph_store
         self.add_class_prefix = add_class_prefix
 
     @classmethod
@@ -32,7 +36,7 @@ class DMSLoader(CDFLoader[dm.InstanceApply]):
     ) -> "DMSLoader":
         # Todo add error handling
         data_model = client.data_modeling.data_models.retrieve(data_model_id, inline_views=True).latest_version()
-        return cls(data_model, graph_store, add_class_prefix)
+        return cls(graph_store, data_model, add_class_prefix)
 
     @classmethod
     def from_rules(
@@ -40,14 +44,15 @@ class DMSLoader(CDFLoader[dm.InstanceApply]):
     ) -> "DMSLoader":
         schema = rules.as_schema()
         # Todo add error handling
-        return cls(schema.as_read_model(), graph_store, add_class_prefix)
+        return cls(graph_store, schema.as_read_model(), add_class_prefix)
 
     def _load(self, stop_on_exception: bool = False) -> Iterable[dm.InstanceApply | NeatValidationError]:
         classes = (view.external_id for view in self.data_model.views)
-        for class_name, triples in self._iterate_class_triples(classes):
+        for class_name in classes:
             # Some tracking and creation of a structure to do validation
             validation_structure = self._create_validation_structure(class_name)
-            for instance_dict in triples2dictionary(triples).values():
+            triples = self.graph_store.queries.list_instances_of_type(class_name)
+            for instance_dict in _triples2dictionary(triples).values():
                 try:
                     yield self._create_instance(class_name, instance_dict, validation_structure)
                 except NeatValidationError as e:
@@ -76,3 +81,73 @@ class DMSLoader(CDFLoader[dm.InstanceApply]):
                 json.dump(dumped, f, indent=2)
             else:
                 yaml.safe_dump(dumped, f, sort_keys=False)
+
+    def _create_validation_structure(self, class_name: str) -> dict:
+        return {}
+
+    def _create_instance(self, class_name: str, instance_dict: dict, validation_structure: dict) -> dm.InstanceApply:
+        raise NotImplementedError()
+
+
+def _triples2dictionary(
+    triples: Iterable[tuple[URIRef, URIRef, str | URIRef]],
+) -> dict[URIRef, dict[URIRef | str, list[str | URIRef]]]:
+    """Converts list of triples to dictionary"""
+    dictionary: dict[URIRef, dict[URIRef | str, list[str | URIRef]]] = {}
+    for triple in triples:
+        id_, property_, value = _remove_namespace(*triple)  # type: ignore[misc]
+        if id_ not in dictionary:
+            dictionary[id_] = defaultdict(list)
+            dictionary[id_]["external_id"].append(id_)
+
+        dictionary[id_][property_].append(value)
+    return dictionary
+
+
+@overload
+def _remove_namespace(*URI: URIRef | str, special_separator: str = "#_") -> str: ...
+
+
+@overload
+def _remove_namespace(*URI: tuple[URIRef | str, ...], special_separator: str = "#_") -> tuple[str, ...]: ...
+
+
+def _remove_namespace(
+    *URI: URIRef | str | tuple[URIRef | str, ...], special_separator: str = "#_"
+) -> tuple[str, ...] | str:
+    """Removes namespace from URI
+
+    Args
+        URI: URIRef | str
+            URI of an entity
+        special_separator : str
+            Special separator to use instead of # or / if present in URI
+            Set by default to "#_" which covers special client use case
+
+    Returns
+        Entities id without namespace
+
+    Examples:
+
+        >>> _remove_namespace("http://www.example.org/index.html#section2")
+        'section2'
+        >>> _remove_namespace("http://www.example.org/index.html#section2", "http://www.example.org/index.html#section3")
+        ('section2', 'section3')
+    """
+    if isinstance(URI, str | URIRef):
+        uris = (URI,)
+    elif isinstance(URI, tuple):
+        # Assume that all elements in the tuple are of the same type following type hint
+        uris = cast(tuple[URIRef | str, ...], URI)
+    else:
+        raise TypeError(f"URI must be of type URIRef or str, got {type(URI)}")
+
+    output = []
+    for u in uris:
+        try:
+            _ = TypeAdapter(HttpUrl).validate_python(u)
+            output.append(u.split(special_separator if special_separator in u else "#" if "#" in u else "/")[-1])
+        except ValidationError:
+            output.append(str(u))
+
+    return tuple(output) if len(output) > 1 else output[0]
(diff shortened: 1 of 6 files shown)


cognite/neat/graph/loaders/_rdf2dms.py
```diff
@@ -0,0 +1,191 @@
...
+from cognite.neat.rules.models.data_types import _DATA_TYPE_BY_DMS_TYPE
+
+from ._base import CDFLoader
+
+
+class DMSLoader(CDFLoader[dm.InstanceApply]):
+    def __init__(
+        self,
+        graph_store: NeatGraphStoreBase,
+        data_model: dm.DataModel[dm.View],
```
nikokaoja, line 29: This would work if underlaying graph's properties and classes are transformed such that:
Class -> View
properties -> view.property_ids

nikokaoja: and that class_id and property_id are in the same namespace, however RDF is often composed of properties and classes which originate from various namespace. 

Two options:
1. transform original/source graph to same namespace, such that class_id and property_id contain only DMS allowed characters (allow Select queries afterwards)
2. transform graph on query, i.e. use construct query to create "virtual graphs"

cognite/neat/graph/loaders/_rdf2dms.py
```diff
@@ -0,0 +1,191 @@
...
+        self.class_by_view_id = class_by_view_id or {}
+        self.add_class_prefix = add_class_prefix
+
+    @classmethod
+    def from_data_model_id(
+        cls,
+        client: CogniteClient,
+        data_model_id: dm.DataModelId,
+        graph_store: NeatGraphStoreBase,
+        add_class_prefix: bool = False,
```
nikokaoja, line 44: This requires graph to be transformed to force usage of same namespace for both properties and classes (see my comment above)

cognite/neat/graph/loaders/_rdf2dms.py
```diff
@@ -0,0 +1,191 @@
...
+                continue
+            dumped[key].append(item.dump())
+        with filepath.open("w", encoding=self._encoding, newline=self._new_line) as f:
+            if filepath.suffix == ".json":
+                json.dump(dumped, f, indent=2)
+            else:
+                yaml.safe_dump(dumped, f, sort_keys=False)
+
+    def _create_validation_structure(self, view: dm.View) -> type[Model]:
+        field_definitions: dict[str, tuple[type, Any]] = {}
```
nikokaoja, line 97: I see that you here want to split between what is a Node (only attributes about node) and what is a Edge (i.e. only references/links) ? This would double number of queries since you are probably quering:
- only `DatatypeProperties` , which fetch only attributes of node
- only `ObjectProperties`, which fetch edges/connections between nodes

cognite/neat/graph/loaders/_rdf2dms.py
```diff
@@ -0,0 +1,191 @@
...
+        def parse_list(cls, value: Any, info: ValidationInfo) -> list[str]:
+            if isinstance(value, list) and cls.model_fields[info.field_name].annotation is not list:
+                if len(value) == 1:
+                    return value[0]
+                raise ValueError(f"Got multiple values for {info.field_name}: {value}")
+            return value
+
+        validators: dict[str, classmethod] = {"parse_list": field_validator("*", mode="before")(parse_list)}  # type: ignore[dict-item,arg-type]
+
+        return create_model(view.external_id, __validators__=validators, **field_definitions)  # type: ignore[arg-type, call-overload]
```
nikokaoja, line 125: note to myself: this returns in-memory Pydantic model

cognite/neat/graph/stores/_base.py
```diff
@@ -355,8 +355,8 @@ def list_instances_of_type(self, class_uri: URIRef) -> list[ResultRow]:
             List of triples for instances of the given class
         """
         query = (
-            f"SELECT ?instance ?prop ?value "
-            f"WHERE {{ ?instance rdf:type <{class_uri}> . ?instance ?prop ?value . }} order by ?instance "
+            "SELECT ?instance ?prop ?value "
+            f"WHERE {{ ?instance <rdf:type> <{class_uri}> . ?instance ?prop ?value . }} order by ?instance "
```
nikokaoja, line 359:
```suggestion
            f"WHERE {{ ?instance a <{class_uri}> . ?instance ?prop ?value . }} order by ?instance "
```

nikokaoja: Since RDF is not schema driven you might end up having a property for which you expect single value, but instead you have list, meaning you will have extra triples coming in, which probably will lead to breaking of pydantic classes.

cognite/neat/graph/loaders/_rdf2dms.py
```diff
@@ -0,0 +1,191 @@
...
+) -> dict[str, dict[str, list[str]]]:
+    """Converts list of triples to dictionary"""
+    values_by_property_by_identifier: dict[str, dict[str, list[str]]] = defaultdict(lambda: defaultdict(list))
+    for triple in triples:
+        id_, property_, value = _remove_namespace(triple, special_separator=":")
+        values_by_property_by_identifier[id_][property_].append(value)
+    return values_by_property_by_identifier
+
+
+def _remove_namespace(URI: ResultRow, special_separator: str = "#") -> list[str]:
```
nikokaoja, line 151: we have `remove_namespace` in utils

tests/tests_unit/test_graph/test_loaders/test_rdf2dms.py
```diff
@@ -0,0 +1,106 @@
...
+from cognite.neat.graph.stores import MemoryStore
+
+
+@pytest.fixture()
+def car_case() -> MemoryStore:
+    store = MemoryStore(namespace=Namespace("http://example.com/"), prefixes={"ex": "http://example.com/"})
+    store.init_graph()
+    store.add_triples(
+        [
+            (URIRef("ex:Car1"), URIRef("rdf:type"), URIRef("ex:Car")),
```
nikokaoja, line 16: For this example structure above works, as all your classes, properties, and instances are in the same namespace. In addition your classes and properties are using DMS compliant characters.

cognite/neat/graph/loaders/_rdf2dms.py
```diff
@@ -0,0 +1,191 @@
...
+        # Todo add error handling
+        return cls(graph_store, schema.as_read_model(), {}, add_class_prefix)
+
+    def _load(self, stop_on_exception: bool = False) -> Iterable[dm.InstanceApply | NeatValidationError]:
+        for view in self.data_model.views:
+            view_id = view.as_id()
+            # Todo Some tracking and creation of a structure to do validation
+            validation_structure = self._create_validation_structure(view)  # type: ignore[var-annotated]
+            uri_ref = self.class_by_view_id.get(view.as_id(), URIRef(f"{view.space}:{view.external_id}"))
+            triples = self.graph_store.queries.list_instances_of_type(uri_ref)
```
nikokaoja, line 64: note to myself: this will read all triples which are then dump to `_triples2dictionary` to create identifier-property-pairs

cognite/neat/graph/loaders/_rdf2dms.py
```diff
@@ -0,0 +1,191 @@
...
+
+from cognite.neat.graph.stores import NeatGraphStoreBase
+from cognite.neat.rules.issues import NeatValidationError
+from cognite.neat.rules.models import DMSRules
+from cognite.neat.rules.models.data_types import _DATA_TYPE_BY_DMS_TYPE
+
+from ._base import CDFLoader
+
+
+class DMSLoader(CDFLoader[dm.InstanceApply]):
```
nikokaoja, line 25: We should leverage in-memory pydantic models to both create assets/relationships and dms nodes and edges

cognite/neat/graph/stores/_base.py
nikokaoja: This will return you both Attributes of a node and Connections of a node, as in RDF properties are simply predicates of a triple. Alt is to explicitly via ontology which property is of type `DatatypeProperty` (attributes) and which `ObjectProperty` (connections). One can get along with `Filter` to select only properties which hold values:

```sparql
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>

SELECT ?property ?value
WHERE {
    ?subject ?property ?value .
    FILTER(isLiteral(?value))
}
```

or IRIs (connection):

```sparql
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>

SELECT ?property ?value
WHERE {
    ?subject ?property ?value .
    FILTER(isIRI(?value))
}
```

there are ways to construct simpler sparql query, merging lists, etc. , though as written above we are forced to transform source graph to singular namespace with ids of properties and classes that DMS compliant.

Somewhere we need to pay the price :)

nikokaoja (commented):
Let's discuss in more details, in short we have two options:
1. Use Select query, but force graph transformation prior loading
2. Use Construct query, skip transformation prior loading

tests/tests_unit/test_graph/test_loaders/test_rdf2dms.py
doctrino: Doing as agreed, assume 'load ready' store with all relevant in the 'neat' namespace

cognite/neat/graph/loaders/_rdf2dms.py
doctrino: As discussed we assume same namespace

tests/tests_unit/test_graph/test_loaders/test_rdf2dms.py
```diff
@@ -0,0 +1,210 @@
...
+                sources=[
+                    dm.NodeOrEdgeData(
+                        source=CAR_MODEL.views[0].as_id(),
+                        properties={"year": 2020, "color": {"space": instance_space, "externalId": "Blue"}},
+                    )
+                ],
+            ),
+            dm.EdgeApply(
+                space=instance_space,
+                external_id="Car1.make.Toyota",
```
nikokaoja, line 55: I like this form of external_id helps reading edge easy

cognite/neat/graph/stores/_base.py
```diff
@@ -361,3 +362,18 @@ def list_instances_of_type(self, class_uri: URIRef) -> list[ResultRow]:
...
+
+    def triples_of_type(self, rdf_type: str) -> list[tuple[str, str, str]]:
+        """Get all triples of a given type.
+
+        This method assumes the graph has been transformed into the default namespace.
+        """
+        rdf_uri = self.store.namespace[rdf_type]
+        query = (
+            f"SELECT ?instance ?prop ?value "
+            f"WHERE {{ ?instance a <{rdf_uri}> . ?instance ?prop ?value . }} order by ?instance "
```
nikokaoja, line 374:
```suggestion
            f"WHERE {{ ?instance a <{self.store.namespace[rdf_type]}> . ?instance ?prop ?value . }} order by ?instance "
```

cognite/neat/graph/stores/_base.py
```diff
@@ -361,3 +362,18 @@ def list_instances_of_type(self, class_uri: URIRef) -> list[ResultRow]:
         logging.info(query)
         # Select queries gives an iterable of result rows
         return cast(list[ResultRow], list(self.store.query(query)))
+
+    def triples_of_type(self, rdf_type: str) -> list[tuple[str, str, str]]:
+        """Get all triples of a given type.
+
+        This method assumes the graph has been transformed into the default namespace.
+        """
+        rdf_uri = self.store.namespace[rdf_type]
```
nikokaoja, line 371:
```suggestion
```

cognite/neat/graph/stores/_base.py
```diff
@@ -361,3 +362,18 @@ def list_instances_of_type(self, class_uri: URIRef) -> list[ResultRow]:
         logging.info(query)
         # Select queries gives an iterable of result rows
         return cast(list[ResultRow], list(self.store.query(query)))
+
+    def triples_of_type(self, rdf_type: str) -> list[tuple[str, str, str]]:
+        """Get all triples of a given type.
```
nikokaoja, line 367: with this method you are getting all triples defining all instances of a given type, so maybe the method name cloud be `triples_of_type_instances` ?

nikokaoja (approved):

PR #474, opened 2024-05-31
Neat 271 always do prescriptive Ontology importing remove make compliant flag

nikokaoja:
### Improved
- `OWLImporter` is now opinionated and will attempt to make the imported ontology compliant with the rules.


doctrino (approved):
Just a control question. Any issue this can cause?

PR #479, opened 2024-06-06
[NEAT-288] 📖as_read for DMSSchema

doctrino:


cognite/neat/rules/models/dms/_schema.py
```diff
@@ -669,7 +679,124 @@ def referenced_spaces(self, include_indirect_references: bool = True) -> set[str
         return referenced_spaces
 
     def as_read_model(self) -> dm.DataModel[dm.View]:
```
nikokaoja, line 681: as read model is ? for graph loader?

nikokaoja (approved):

PR #482, opened 2024-06-06
[NEAT-289] 👨‍🏫 Tracking ingestion

doctrino:


nikokaoja (approved):

cognite/neat/graph/_tracking/base.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC, abstractmethod
+from collections.abc import Sequence
+
+from cognite.neat.issues import NeatIssue
+
+
+class Tracker(ABC):
```
alivinco, line 7: We also might need a method for retrieving current trackable state , something like get() method. In log tracker it would be value of the counter.

doctrino: Lets add it when needed.

PR #484, opened 2024-06-06
[NEAT-290] 📠 DMS Loader to CDF

doctrino:
**Note** Most of this PR is moving things around.

* Moved car example out for reuse in integration test.
* Moved `_model.py` to `utils/upload.py` for reuse in `graph`.
* Implemented upload for `DMSLoader`.

**Note**: I did not implement any batching or retry logic. I know that there has been an effort in the python-SDK to do this. So I think we should try to avoid doing it as well and only do it if we start to see issues.

image


cognite/neat/graph/loaders/_rdf2dms.py
```diff
@@ -7,8 +8,10 @@
 import yaml
 from cognite.client import CogniteClient
 from cognite.client import data_modeling as dm
+from cognite.client.data_classes.capabilities import Capability, DataModelInstancesAcl
```
nikokaoja, line 11: what is this? :)

cognite/neat/utils/upload.py
```diff
@@ -64,3 +64,23 @@ def as_report_str(self) -> str:
             line.append(f"failed to delete {self.failed_deleted}")
 
         return f"{self.name.title()}: {', '.join(line)}"
+
+
```
nikokaoja, line 68: nica

nikokaoja (approved):

cognite/neat/graph/loaders/_rdf2dms.py
doctrino: It checks that the client that is passed have authorization to write to CDF data model instances.

PR #489, opened 2024-06-07
added provo framework

nikokaoja:
Added PROV-O framework for keeping track of graph provenance.


doctrino (commented):
New concept for me to learn :) I like it sems to be the right theory for the practical implementation. 

If I understand correctly, the Extractors will then have a `.as_change` method which returns an instance of the `Change` class?

doctrino (approved):

PR #494, opened 2024-06-11
Neat 300 extractor dexpi

nikokaoja:
### Added
- `DexpiExtractor` graph extractor added.


cognite/neat/graph/stores/_base.py
```diff
@@ -124,11 +124,22 @@ def from_oxi_store(cls, storage_dir: Path | None = None, rules: InformationRules
         return cls(graph, rules)
 
     def write(self, extractor: TripleExtractors) -> None:
+        _start = datetime.now(pytz.utc)
```
doctrino, line 127: Maybe worth a discussion. Do we want have pytz as a dependency for neat? Even though it is useful, as neat is a python package, we should strive to have as few dependencies as possible.

cognite/neat/graph/extractors/_dexpi.py
```diff
@@ -0,0 +1,202 @@
...
+        filepath: Path,
+        namespace: Namespace | None = None,
+    ):
+        self.filepath = filepath
+        self.namespace = namespace or DEFAULT_NAMESPACE
+        self.root = ET.parse(self.filepath).getroot()
+
+    @classmethod
+    def from_file(cls, file_path: str | Path, namespace: Namespace | None = None):
+        return cls(Path(file_path), namespace)
```
doctrino, line 38: I suggest the following:
```suggestion
    def __init__(
        self,
        filepath: Path,
        namespace: Namespace | None = None,
    ):
        self.filepath = filepath
        self.namespace = namespace or DEFAULT_NAMESPACE
        self.root = ET.parse(self.filepath).getroot()

    @classmethod
    def from_file(cls, file_path: str | Path, namespace: Namespace | None = None):
        return cls(Path(file_path), namespace)
```
```suggestion
    def __init__(
        self,
        root: Element,
        namespace: Namespace | None = None,
    ):
        self.root = root
        self.namespace = namespace or DEFAULT_NAMESPACE

    @classmethod
    def from_file(cls, file_path: str | Path, namespace: Namespace | None = None):
        return cls(ET.parse(file_path).getroot(), namespace)
```

As far as it does not hurt performance I think extractors should take the core data object in the `__init__` method that it works on the `.extract` method. 

Then, if we want to extend it in the future to for example read a dexpi from an URL it is easy to extend it with
```python
    @classmethod
    def from_url(cls, url: str, namespace: Namespace | None = None):
        import requests
        from io import BytesIO

        response = requests.get(url)
        response.raise_for_status()
        return cls(ET.parse(BytesIO(response.content)).getroot(), namespace)
```
With the current implementation the above would not be possible and you would have to dump a file to disk and read it again.

cognite/neat/graph/extractors/_dexpi.py
```diff
@@ -0,0 +1,202 @@
...
+                        id_,
+                        DEXPI[f"children/{camel_case_property}"],
+                        namespace[child.attrib["ID"]],
+                    )
+                )
+
+        return triples
+
+    @classmethod
+    def _to_uri_friendly_association_type(cls, association):
```
doctrino, line 111: type hints

cognite/neat/graph/extractors/_dexpi.py
```diff
@@ -0,0 +1,202 @@
...
+        if children := get_children(element, "GenericAttributes", 1):
+            if grandchildren := get_children(children[0], "GenericAttribute"):
+                for generic_attribute in grandchildren:
+                    # extension for schema version 3.3, where "AttributeURI" is not included
+                    if "AttributeURI" in generic_attribute.attrib:
+                        if generic_attribute.attrib["AttributeURI"] not in attributes:
+                            attributes[generic_attribute.attrib["AttributeURI"]] = [generic_attribute.attrib]
+
+                        else:
+                            attributes[generic_attribute.attrib["AttributeURI"]].append(generic_attribute.attrib)
```
doctrino, line 200: Good case for `defaultdict(list)`?
```suggestion
        attributes = defaultdict(list)
        if children := get_children(element, "GenericAttributes", 1):
            if grandchildren := get_children(children[0], "GenericAttribute"):
                for generic_attribute in grandchildren:
                    # extension for schema version 3.3, where "AttributeURI" is not included
                    if "AttributeURI" in generic_attribute.attrib:
                        attributes[generic_attribute.attrib["AttributeURI"]].append(generic_attribute.attrib)

        return dict(attributes)
```

doctrino (approved):
Great job!

A few suggestion and one point up for discussion wrt to the `__init__` method.

cognite/neat/graph/stores/_base.py
nikokaoja: Agree, but it is not our direct dep. we can though remove it from usage.

nikokaoja: Fixed

cognite/neat/graph/extractors/_dexpi.py
nikokaoja: I like this!

cognite/neat/graph/extractors/_dexpi.py
nikokaoja: yeps :)

PR #500, opened 2024-06-14
Neat 307 dexpi extractor handling rdf type for componenets without component class uri

nikokaoja:
### Improved
- Dexpi Extractor is significantly more extracting triples from Dexpi XML files
- More human readable class and property ids in Dexpi Extractor


cognite/neat/graph/extractors/_dexpi.py
```diff
@@ -13,6 +13,7 @@
 from cognite.neat.utils.xml import get_children, iterate_tree
 
 DEXPI = Namespace("http://sandbox.dexpi.org/rdl/")
+QUDT = Namespace("https://qudt.org/vocab/unit/")
```
doctrino, line 16: Do we need unit support to properly support this?

doctrino (approved):

PR #508, opened 2024-06-20
Neat 318 add query module to graph

nikokaoja:
### Added
- Introduce `query` module under `neat.graph` which holds previous `_Queries` class
- Added generation of `SPARQL` `CONSTRUCT` queries based on `rdfpath` transformations defined in `InformationRules`
- Introduce `NeatGraphStore.read` method which takes class and returns all instances of that class
- Test for `NeatGraphStore.read` method which entails end-to-end process of loading triples, inferring data model and reading instances of a class
### Changed
- `DMSLoader` now uses `.read` method of `NeatGraphStore`


doctrino (approved):
If I understand this PR correctly, this is mainly moving code around, it is not much new. 

The only thing that reacted to, should the `queries` package be inside the `store`? As it is mostly a helper, and not a core concept, like `extractors`, `store`, `loader`, or `transformer`

PR #516, opened 2024-06-26
Neat 333 query build asset centric construct query

nikokaoja:
### Added
- Conversion between information, asset and dms rules
- Added serializer for transformations (i.e. RDFPATH)
- Placeholder for AssetLoader


cognite/neat/graph/stores/_base.py
```diff
@@ -64,11 +64,11 @@ def __init__(
 
         self.queries = Queries(self.graph, self.rules)
 
-    def add_rules(self, rules: InformationRules) -> None:
+    def add_rules(self, rules: InformationRules | AssetRules) -> None:
```
doctrino, line 67: Not sure if I agree with this one. Do we allow implementation rules in the Store?

doctrino (approved):
Like the nice conversion. I am a bit confused by the AssetRules in the NeatStore.

PR #518, opened 2024-06-26
[NEAT-337]😌 Interface Loaders

doctrino:
Simplification


cognite/neat/graph/loaders/_base.py
```diff
@@ -88,10 +60,11 @@ def _load_into_cdf_iterable(
                 items.append(result)
 
             if len(items) >= self._UPLOAD_BATCH_SIZE:
-                yield self._upload_to_cdf(client, items, return_diffs, dry_run, issues)
-                items.clear()
+                yield self._upload_to_cdf(client, items, dry_run, issues)
+                issues = NeatIssueList[NeatIssue]()
+                items = []
```
doctrino, line 65: This was a bug, we cleared the list of issues already returned to the user. In addition, we would include all issuse found in the entire run and not just the ones on the current batch.

nikokaoja (approved):

PR #522, opened 2024-06-27
Push image to f25e common container registry

paramids:


pavelzubarev (approved):

aastor (approved):

.github/workflows/docker-builder-publisher.yaml
```diff
@@ -30,14 +30,27 @@ jobs:
         run: |
           poetry export -f requirements.txt --output requirements.txt --extras "all"
 
+      - id: version
+        name: Build Version
+        run: echo "version=$(date -u '+%Y%m%dT%H%M%SZ')"  >> $GITHUB_ENV
```
doctrino, line 35: Why introduce another type of versioning that the image pushed to docker hub?

doctrino (approved):

.github/workflows/docker-builder-publisher.yaml
paramids: f25e we use timestamp for the image versions

PR #525, opened 2024-06-28
Sol 866: Add permissions section to build job

paramids:


doctrino (approved):

.github/workflows/docker-builder-publisher.yaml
```diff
@@ -37,7 +41,7 @@ jobs:
       - name: Azure login
         uses: azure/login@v1
         with:
-          client-id: 9c8d1e31-54b5-4156-b1d0-408135e2b55b
+          client-id: 7e26ba04-3da0-4241-b517-c12b8205af8a
```
pavelzubarev, line 44: I have not checked that ID, but it should be the one you have terraformed federation for.

.github/workflows/docker-builder-publisher.yaml
```diff
@@ -11,6 +11,10 @@ jobs:
   build-and-publish:
     runs-on: ubuntu-latest
     environment: CD
```
pavelzubarev, line 13: federation may requite environment to be called `main`

pavelzubarev (approved):
Approved with comments.

.github/workflows/docker-builder-publisher.yaml
pavelzubarev: You may need to create new action that build image independently into our internal registries that has `main` as an environment

.github/workflows/docker-builder-publisher.yaml
```diff
@@ -30,17 +30,49 @@ jobs:
...
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+        with:
+          version: 1.3.1
```
pavelzubarev, line 59: update that to the latest, if possible

paramids: updated to latest release version 1.3.4

PR #529, opened 2024-07-06
Relax version requirement for cognite-sdk

mortenmj:
The current version requirement hinders updates in the monorepo. Hopefully this works.


doctrino (changes requested):
We cannot do that. Neat depends on the lower bound for the SDK as it uses the newest features of it, if we lower it like you suggest neat might stop working for users.

PR #536, opened 2024-07-09
Neat 355 inference importer json string value type

nikokaoja:
### Added
- Option for checking for JSON value type when inferring data model


doctrino (approved):
I realize now that you made not made unpacking the metadata the default. I think metadata should be unpacked by default. The reason is that the most basic case (least skilled user) this is the desired behavior.

PR #541, opened 2024-07-15
READ via DESCRIBE query

nikokaoja:
### Changed
- `NeatGraphStore.read()` is now iterable utilizing `DESCRIBE` query instead of `CONSTRUCT` query.
### Improved
- Order of magnitude improved query speed of instances for downstream graph loaders


cognite/neat/utils/utils.py
```diff
@@ -75,15 +87,25 @@ def _get_cognite_client(config: CogniteClientConfig, credentials: CredentialProv
...
+    special_separator: str = "#_",
+    deep_validation: bool = False,
+) -> tuple[str, ...]: ...
 
 
 def remove_namespace_from_uri(
-    *URI: URIRef | str | tuple[URIRef | str, ...], special_separator: str = "#_"
+    *URI: URIRef | str | tuple[URIRef | str, ...],
+    special_separator: str = "#_",
+    deep_validation: bool = False,
```
doctrino, line 108: I think it is good to avoid flags if we can.
What about making this a Literal and then in the docs explain the difference. 
```suggestion
    validation: Literal["full", "prefix"] = "prefix",
```
```python
"""
validation: How to validate the URI. Full checks that it is a valid HttpURL, while prefix that it starts with 'http'
"""
```

cognite/neat/utils/utils.py
```diff
@@ -192,7 +220,9 @@ def _traverse(hierarchy: dict, graph: dict, names: list[str]) -> dict:
 
 
 def get_generation_order(
-    class_linkage: pd.DataFrame, parent_col: str = "source_class", child_col: str = "target_class"
+    class_linkage: pd.DataFrame,
+    parent_col: str = "source_class",
+    child_col: str = "target_class",
```
doctrino, line 225: Is this a preference? Or do you have some sort of plugin that changes the style?

Not a big deal, but all these irrelevant multiline changes makes the PR a bit more difficult to review.

doctrino (approved):

PR #547, opened 2024-07-17
Neat 369 describe query improve renaming of properties

nikokaoja:
### Added
- `AddAllReferences` transformer that handles `AllReferences` RDF path in Rules
### Improved
- Better handling of property renaming in `DESCRIBE` query, which for example allows RDF:type property to be used
- Iterating over classes which have properties defined for them instead of all classes (which causes errors)
### Removed
- `AllProperties` RDF path has been removed as we no longer want support implicit properties


cognite/neat/graph/stores/_base.py
```diff
@@ -179,17 +179,27 @@ def read(self, class_: str) -> Iterable[tuple[str, dict[str, list[str]]]]:
             warnings.warn("Desired type not found in graph!", stacklevel=2)
             return None
 
-        if InformationAnalysis(self.rules).has_hop_transformations():
+        if (has_hop_transformations := InformationAnalysis(self.rules).has_hop_transformations()) or (
+            has_all_reference_transformations := InformationAnalysis(self.rules).has_all_reference_transformations()
+        ):
```
doctrino, line 184: See comment in last PR.

cognite/neat/graph/transformers/_rdfpath.py
```diff
@@ -1,7 +1,49 @@
...
 
 
 class ReduceHopTraversal(BaseTransformer):
     """ReduceHopTraversal is a transformer that reduces the number of hops to direct connection."""
 
     ...
+
+
+class AddAllReferences(BaseTransformer):
+    description: str = "Adds all references to specific property defined in Rules"
```
doctrino, line 17: See comment in last PR.

PR #556, opened 2024-07-21
[NEAT-378] 💣 Removed superfluous linters

doctrino:
Removed all linters that were already checked by ruff.


cognite/neat/app/api/routers/rules.py
```diff
@@ -166,7 +166,7 @@ def get_original_rules_from_workflow(workflow_name: str):
         return {"error": f"Workflow {workflow_name} is not found"}
     context = workflow.get_context()
     rules_data = context["RulesData"]
-    if type(rules_data) != RulesData:
+    if type(rules_data) is not RulesData:
```
doctrino, line 169: upgraded ruff requires this (it is correct to compare memory location of types as types are singletons).

nikokaoja (approved):

PR #563, opened 2024-07-23
[NEAT-353] 😯Remove regex from inference. 

doctrino:
image


nikokaoja (approved):
great that usage of f strings!

PR #566, opened 2024-07-23
[NEAT-363] 🔥 Speedup tests

doctrino:
* Switch to oxy graph
* Some code speedups
* Deleted unused test data.

Overall 1m12s to 18-21s on my machine (only unit tests)

I skipped parallelization as 18s is so fast that the overhead of extra workers is unlikely to pay off.


tests/tests_unit/rules/test_importers/test_inference_importer.py
```diff
@@ -8,13 +8,20 @@
...
 
     assert len(rules.properties) == 312
     assert len(rules.classes) == 59
 
     # checking multi-value type
-    assert set(rules.properties.data[19].value_type.types) == set(
+    multi_value_property = "OperatingShare.PowerSystemResource"
+    prop = next((prop for prop in rules.properties if prop.property_ == multi_value_property), None)
+    assert prop is not None, f"Failed to infer expected multi-value property {multi_value_property}"
+    assert set(prop.value_type.types) == set(
```
doctrino, line 24: oxy seems to be not be deterministic, so need a most robust check here

cognite/neat/exceptions.py
```diff
@@ -102,6 +102,9 @@ def __str__(self) -> str:
         return self.message
 
 
+class NeatValueError(NeatException, ValueError): ...
```
doctrino, line 105: It is good practice to add our own exception were it is a user mistake. Then we can later have for example sentry catch all non Neat error which would be bugs in the code.

cognite/neat/rules/importers/_spreadsheet2rules.py
```diff
@@ -131,25 +131,24 @@ def sheet_names(self, role: RoleTypes) -> set[str]:
...
-                metadata = self.metadata
-            else:
-                metadata = self._read_metadata(excel_file, filepath)
-                if metadata is None:
-                    # The reading of metadata failed, so we can't continue
-                    return None
-
-            sheets, read_info_by_sheet = self._read_sheets(excel_file, metadata.role)
-            if sheets is None or self.issue_list.has_errors:
+    def read(self, excel_file: pd.ExcelFile, filepath: Path) -> None | ReadResult:
```
doctrino, line 134: Switch this one to opening the excel file outside. It is faster than opening the excel files for each set of sheets we want to read.

tests/tests_unit/test_neat_package_metadata.py
```diff
@@ -35,6 +36,16 @@ def test_consistent_version_variables():
...
 
 def test_no_spaces_in_sub_folders() -> None:
-    name_by_location = {path: path.name for path in (ROOT / "cognite").rglob("*") if path.is_dir() and " " in path.name}
+    name_by_location: dict[Path, str] = {}
+    to_check = [ROOT / "cognite"]
+    while to_check:
+        current = to_check.pop()
+        for path in current.iterdir():
+            if path.is_dir():
+                if path.name not in {"node_modules"}:
```
doctrino, line 45: Avoid searching the `node_modules` shaves off 15 seconds on this test.

cognite/neat/graph/queries/_base.py
```diff
@@ -98,55 +98,48 @@ def describe(
...
 
         Returns:
             Dictionary of instance properties
         """
-
         property_values: dict[str, list[str]] = defaultdict(list)
-
-        for subject, predicate, object_ in cast(list[ResultRow], self.graph.query(f"DESCRIBE <{instance_id}>")):
-            if object_.lower() not in [
+        identifier = remove_namespace_from_uri(instance_id, validation="prefix")
```
doctrino, line 112: The subject will be the same for each iteration, so only calling `remove_namespace_uri` once on it save some time

cognite/neat/utils/rdf_.py
```diff
@@ -10,25 +12,28 @@
...
 
 def remove_namespace_from_uri(
-    *URI: URIRef | str | tuple[URIRef | str, ...],
+    URI: URIRef | str | SequenceNotStr[URIRef | str],
+    *,
     special_separator: str = "#_",
     validation: Literal["full", "prefix"] = "prefix",
-) -> tuple[str, ...] | str:
+) -> str | list[str]:
     """Removes namespace from URI
```
doctrino, line 37: The overload was not correctly implemented leading to some unnecessary casting. So fixed the implementation

cognite/neat/graph/stores/_base.py
```diff
@@ -221,7 +223,7 @@ def _parse_file(
         Args:
             filepath : File path to file containing graph data, by default None
             mime_type : MIME type of graph data, by default "application/rdf+xml"
```
nikokaoja, line 225: Revert to previous as base_uri is not always needed to load file to oxigraph. It is only needed when RDF file contains relative URIs, which is the case with Nordic44. In majority of cases there are absolute URI.

tests/tests_unit/rules/test_importers/test_inference_importer.py
nikokaoja: The order of processing ?

doctrino: Running this tests a few times the "OperatingShare.PowerSystemResource" property is not always on index 19, it changes from run to run.

cognite/neat/graph/stores/_base.py
doctrino: Got it, will revert this change.

tests/tests_unit/rules/test_importers/test_inference_importer.py
nikokaoja: that is expected, I think with in-memory rdflib store they do indexing of triples or something similar to maintain the order

nikokaoja (approved):

PR #568, opened 2024-07-24
[NEAT-354] 🙂‍↕️Shared regex in utility

doctrino:
This is part 1 of cleaning up warnings and exceptions.


cognite/neat/utils/regex_patterns.py
```diff
@@ -0,0 +1,52 @@
+import re
+from functools import cached_property
+
+MORE_THAN_ONE_NONE_ALPHANUMERIC_REGEX = r"([_-]{2,})"
```
nikokaoja, line 4: this is great, as we are having them at couple of places, at least we know now where to go to find what we need, and/or add what is missing !

nikokaoja (approved):

PR #571, opened 2024-07-24
feat: Add IMF rules importer (#552)

nikokaoja:
* feat: Add IMF rules importer

A ned step for importing rules has been added. The step converts Information Modelling Framework (IMF) types, given as SHACL shapes, into rules.

The new importer follows the compliance rules set by the OWL Importer, but most of the logic are, in this case, included the SPARQL queries. However, the validation checks are kept to ensure conformance.

The IMF identifiers are given as UUIDs, but CDF does not allow identifiers starting with a number. All the relevant identifiers are therefore prefixed with "IMF_"

IMF - attributes are properties that are described with several triples. An attribute contains relations to reference data describing the meaning of the attribute, as well as potential qualifiers and units of measure. To ensure that this information is included the attributes are created as both a class and property. This punning is required to ensure that both the usage of the attribute relative to an IMF block or terminal and the relationship to relevant reference data is captured.

* fix: Update value type

In cases where a sh:property points to a sh:class or a sh:qualifiedValueShape the corresponding value is set as the expected value type of the property.

The logic should also cover rdfs:range.

---------

### DIFF
--- a/cognite/neat/rules/importers/__init__.py
+++ b/cognite/neat/rules/importers/__init__.py
@@ -1,6 +1,7 @@
 from ._base import BaseImporter
 from ._dms2rules import DMSImporter
 from ._dtdl2rules import DTDLImporter
+from ._imf2rules import IMFImporter
 from ._inference2rules import InferenceImporter
 from ._owl2rules import OWLImporter
 from ._spreadsheet2rules import ExcelImporter, GoogleSheetImporter
@@ -9,6 +10,7 @@ from ._yaml2rules import YAMLImporter
 __all__ = [
     "BaseImporter",
     "OWLImporter",
+    "IMFImporter",
     "DMSImporter",
     "ExcelImporter",
     "GoogleSheetImporter",
--- a/cognite/neat/rules/importers/_imf2rules/__init__.py
+++ b/cognite/neat/rules/importers/_imf2rules/__init__.py
@@ -1,0 +1,3 @@
+from ._imf2rules import IMFImporter
+
+__all__ = ["IMFImporter"]
--- a/cognite/neat/rules/importers/_imf2rules/_imf2classes.py
+++ b/cognite/neat/rules/importers/_imf2rules/_imf2classes.py
@@ -1,0 +1,240 @@
+from typing import cast
+
+import numpy as np
+import pandas as pd
+from rdflib import OWL, Graph
+
+from cognite.neat.rules.models._base import MatchType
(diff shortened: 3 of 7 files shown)


cognite/neat/rules/importers/_rdf/_imf2rules/_imf2rules.py
```diff
@@ -4,25 +4,26 @@
...
 
 
-class OWLImporter(BaseImporter):
-    """Convert OWL ontology to tables/ transformation rules / Excel file.
+class IMFImporter(BaseImporter):
+    """Convert SHACL shapes to tables/ transformation rules / Excel file.
 
         Args:
-            filepath: Path to OWL ontology
+            filepath: Path to RDF file containing the SHACL Shapes
```
doctrino, line 23: Should this take in filepath? It looks like you immediately start up a graph. So suggest that the init takes a Graph, or a NeatStore, and then have a factory method for this.

cognite/neat/rules/importers/_rdf/_imf2rules/_imf2rules.py
```diff
@@ -61,11 +62,14 @@ def to_rules(
...
         components = {
-            "Metadata": parse_owl_metadata(graph),
-            "Classes": parse_owl_classes(graph),
-            "Properties": parse_owl_properties(graph),
+            "Metadata": parse_imf_metadata(),
+            "Classes": parse_imf_to_classes(graph),
+            "Properties": parse_imf_to_properties(graph),
         }
 
         components = make_components_compliant(components)
```
doctrino, line 75: Below there is no catching of errors if the InformationRules failes

cognite/neat/rules/importers/_rdf/_shared.py
```diff
@@ -0,0 +1,586 @@
...
+
+def clean_up_properties(df: pd.DataFrame) -> pd.DataFrame:
+    class_grouped_dfs = df.groupby("Class")
+
+    clean_list = []
+
+    for class_, class_grouped_df in class_grouped_dfs:
+        property_grouped_dfs = class_grouped_df.groupby("Property")
+        for property_, property_grouped_df in property_grouped_dfs:
+            clean_list += [
```
doctrino, line 224: Should there be warnings here if there are multiple values for any of the properties or is that not possible?

doctrino (approved):
A few concerns around dealing with errors and warnings, but that can be moved to a future maintenance task.

Most important thing to address is whether the init signature is correct and whether to introduce factory methods.

PR #586, opened 2024-08-20
fix imf import error

nikokaoja:
### Fixed
- IMF rules importer failing due to references
### Improved
- Handling of references for OWL importer
### Added
- Test for IMF importer


doctrino (approved):
I was a bit shocked by the +991 to fix a bug, I see mostly test data :)

PR #590, opened 2024-08-24
[NEAT-403] 🚃 Verification transformers

doctrino:
## Discussion
Changed the `RulesTransformers` to be Monads, meaning they can either return a rule or maybe return a rule. The motivation is to handle verification which might fail. More about Monads: https://www.youtube.com/watch?v=Q0aVbqim5pE

The basic idea is that all `RulesTransformers` returns an `OutRules` object which is a wrapper abound a `Rules` object. For example, `ConversionTransformer`, will return the object `JustRules` as conversion always returns just rules. While the `VerificationTransformers` will return a `MaybeRules` object indicating that a verification will maybe return a rules an subsequent transformers should handle whether a rules object is there or not.


cognite/neat/issues/errors/_general.py
```diff
@@ -10,6 +10,13 @@ class NeatValueError(NeatError, ValueError):
     raw_message: str
 
 
+@dataclass(frozen=True)
+class NeatTypeError(NeatError, TypeError):
+    """{raw_message}"""
+
+    raw_message: str
```
doctrino, line 17: I had forgotten this generic type of errors, so added it now.

cognite/neat/rules/transformers/_base.py
```diff
@@ -1,24 +1,83 @@
...
-    def transform(self, rules: T_RulesIn) -> T_RulesOut:
+    def transform(self, rules: T_RulesIn | OutRules[T_RulesIn]) -> OutRules[T_RulesOut]:
         """Transform the input rules into the output rules."""
         raise NotImplementedError()
 
+    @classmethod
+    def _to_rules(cls, rules: T_RulesIn | OutRules[T_RulesIn]) -> T_RulesIn:
+        if isinstance(rules, JustRules):
+            return rules.rules
+        elif isinstance(rules, MaybeRules):
```
nikokaoja, line 58: Would we have extra information why mayberules are None?
Something to discuss on Wed.

cognite/neat/rules/transformers/_converters.py
```diff
@@ -38,34 +40,49 @@
...
+T_VerifiedOutRules = TypeVar("T_VerifiedOutRules", bound=VerifiedRules)
 
-class InformationToDMS(RulesTransformer[InformationRules, DMSRules]):
+
+class ConversionTransformer(RulesTransformer[T_VerifiedInRules, T_VerifiedOutRules], ABC):
+    """Base class for all conversion transformers."""
+
+    def transform(self, rules: T_VerifiedInRules | OutRules[T_VerifiedInRules]) -> JustRules[T_VerifiedOutRules]:
+        out = self._transform(self._to_rules(rules))
+        return JustRules(out)
```
nikokaoja, line 54: Note to myself: conversion must always succeeded , so `JustRules` is our return type

cognite/neat/rules/transformers/_verification.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def transform(self, rules: T_InputRules | OutRules[T_InputRules]) -> MaybeRules[T_VerifiedRules]:
+        issues = IssueList()
+        in_: T_InputRules = self._to_rules(rules)
+        error_args: dict[str, Any] = {}
+        if isinstance(rules, ReadRules):
+            error_args = rules.read_context
+        verified_rules: T_VerifiedRules | None = None
+        with _handle_issues(issues, NeatError, NeatWarning, error_args) as future:
+            verified_rules = self._rules_cls.model_validate(in_.dump())  # type: ignore[assignment]
```
nikokaoja, line 39: We moved _rules_cls all the validation methods inclluding post validation that runs on the entire class

nikokaoja (approved):

cognite/neat/rules/transformers/_converters.py
doctrino: That is absolutely up for discussion.

PR #592, opened 2024-08-24
[NEAT-402] 🗺 Mapping transformer

doctrino:


cognite/neat/rules/_shared.py
```diff
@@ -1,19 +1,61 @@
...
 T_VerifiedRules = TypeVar("T_VerifiedRules", bound=VerifiedRules)
+T_InputRules = TypeVar("T_InputRules", bound=InputRules)
+
+
+@dataclass
+class OutRules(Generic[T_Rules], ABC):
+    """This is a base class for all rule states."""
+
+    @abstractmethod
+    def get_rules(self) -> T_Rules | None:
```
nikokaoja, line 31: why is this method needed?

nikokaoja (approved):

cognite/neat/rules/_shared.py
doctrino: To have a way to access rules from a generic `OutRules` object. I did not move `rules` up as it has different type hints in `JustRules` and `MaybeRules`.

PR #593, opened 2024-08-24
[NEAT-405]🧭Refactor Input Rules

doctrino:
Introduces
* Shared base classes for InputRules
* Tests to ensure up-to-date with Pydantic classes.
* Some code reduction due to the use of base class

Goal: Consistency in InputClasses along with easier to maintain.


cognite/neat/rules/models/entities.py
```diff
@@ -615,3 +616,46 @@ def _generate_cdf_resource_list(v: Any) -> list[AssetEntity | RelationshipEntity
     AnyHttpUrl,
     PlainSerializer(lambda v: str(v), return_type=str, when_used="unless-none"),
 ]
+
+
+def load_value_type(
```
nikokaoja, line 621: maybe rename it to load_information_value_type to be consistent with naming below

nikokaoja (approved):
nice reduction in the code base

PR #595, opened 2024-08-28
fix: update identities created by IMF-importer

bhtvedt:
Main fix:
The IMF identifiers are GUIDs. In the original feature these identifiers were prefixed with IMF_ to meet the requirement that the identifiers shouldn't begin with a number.

An additional requirement is that the identifiers don't contain hypens. In this commit the hypens are replaced by underscore.

Minor updates:
- Attributes are types both DatatypeProperty and AttributeType. This caused the previous version to create duplicate entries for the attributes, one with propertyType DatatypeProperty and the other with ObjectProperty.

The SPARQL has been updated to remove the duplication.

- imf:Attributes are objects, i.e. they have their own triples carrying information about their meaning. The main predicate linked to an imf:AttributeType is the imf:predicate, which points to a base reference term. The imf:predicate is required to create an attribute, and the cardinality was therefore set to 1-1. However, when the spreadsheet was reviewed the meaning became that a block required the attribute. The latter interpretation is not correct and therefore the cardinality of the imf:predicate has been returned to the default 0-1.

- A bug in PCA has lead to the use of the non-existing skos:description term instead of skos:definition. This will be fixed in PCA, and therefore support for both terms have been added in the IMF-importer. The support for skos:description can be removed at a later stage.


cognite/neat/rules/importers/_rdf/_imf2rules/_imf2classes.py
```diff
@@ -37,16 +37,20 @@ def parse_imf_to_classes(graph: Graph, language: str = "en") -> list[dict]:
...
+        # Finding IMF - elements
         VALUES ?type { imf:BlockType imf:TerminalType imf:AttributeType }
         ?imfClass a ?type .
         OPTIONAL {?imfClass rdfs:subClassOf ?parent }.
         OPTIONAL {?imfClass rdfs:label | skos:prefLabel ?name }.
-        OPTIONAL {?imfClass rdfs:comment | skos:description ?description} .
+
+        # Note: Bug in PCA has lead to the use non-existing term skos:description. This will be replaced
+        # with the correct skos:definition in the near future, so both terms are included here.
+        OPTIONAL {?imfClass rdfs:comment | skos:definition | skos:description ?description} .
```
nikokaoja, line 48: smart to handle various way of expressing description via "OR"

nikokaoja (approved):

nikokaoja:
Make sure to update version (bug fix) and update CHANGELOG

PR #596, opened 2024-08-30
[NEAT-296] 🐬 Added nodes

doctrino:
No verification is included. Lets save this for later as it is easy to get wrong and thus block the users.

image


cognite/neat/rules/models/dms/_rules_input.py
```diff
@@ -216,12 +217,34 @@ def from_view(cls, view: dm.ViewApply, in_model: bool) -> "DMSInputView":
         )
 
 
+@dataclass
+class DMSInputNode(InputComponent[DMSNode]):
+    node: str
+    usage: Literal["type", "collocation"]
```
nikokaoja, line 223: isn't it collection instead of collocation ?

nikokaoja (approved):

PR #598, opened 2024-08-31
[prep NEAT-260] 🐙 Refactor Entities

doctrino:
Main motivation; Ensure that the you can import an `Entity` into the data types without a circular import. In the process, cleaned  up and split the large entity file to make it easier to work with.


cognite/neat/rules/models/dms/_rules.py
```diff
@@ -38,13 +38,15 @@
     DMSNodeEntity,
     DMSUnknownEntity,
     EdgeEntity,
+    HasDataFilter,
+    NodeTypeFilter,
+    RawFilter,
```
nikokaoja, line 43: Not related to this PR but wandering are we capturing arbitrary filters (raw filters) when doing DMSImport ?

cognite/neat/rules/models/entities/_multi_value.py
```diff
@@ -0,0 +1,67 @@
+from typing import Any, ClassVar
```
nikokaoja, line 1: I guess this is to make some distinction between previous _base, which is turned into _signel_value.py

nikokaoja (approved):
Nice clean up and update.

Maybe at naming of entities modules is a bit confusing (specifically single_value, multi_value), but I have no good suggestion how to name them. Nevertheless, as these are private methods, not be exposed to end-user, I think this is fine as it is.

PR #609, opened 2024-09-05
[NEAT-93] 🥳 DMS Extractor

doctrino:


cognite/neat/graph/extractors/_dms.py
```diff
@@ -0,0 +1,162 @@
...
+                    yield id_, namespace[key], object_
+
+    def _get_objects(self, value: PropertyValue) -> Iterable[Literal | URIRef]:
+        if isinstance(value, str | float | bool | int):
+            yield Literal(value)
+        elif isinstance(value, dict) and "space" in value and "externalId" in value:
+            yield self._as_uri_ref(dm.DirectRelationReference.load(value))
+        elif isinstance(value, dict):
+            # This object is a json object.
+            yield Literal(str(value))
```
doctrino, line 124: @nikokaoja Need some input here

cognite/neat/graph/loaders/_rdf2dms.py
```diff
@@ -108,7 +110,7 @@ def _load(self, stop_on_exception: bool = False) -> Iterable[dm.InstanceApply |
         for view in self.data_model.views:
             view_id = view.as_id()
             tracker.start(repr(view_id))
-            pydantic_cls, edge_by_properties, issues = self._create_validation_classes(view)  # type: ignore[var-annotated]
+            pydantic_cls, edge_by_type, issues = self._create_validation_classes(view)  # type: ignore[var-annotated]
```
doctrino, line 113: Edges are identified by type, it was a mistake to use the property id in the schema. That is just a the name used in the schema, which might not be in the triples at all.

cognite/neat/graph/extractors/_dms.py
nikokaoja: Why we are doing this? The reason being that when we are loading this to Nodes it will be properly resolved as JSON (i.e. python dict), otherwise you lose datatype and property value gets provided as string.

nikokaoja: here in loader: https://github.com/cognitedata/neat/blob/ef613bbee327b26d113ba3793654225a810f4702/cognite/neat/graph/loaders/_rdf2dms.py#L198

nikokaoja: Also helps with inference of data model

cognite/neat/graph/queries/_base.py
```diff
@@ -126,7 +127,8 @@ def describe(
             # losing the namespace from the predicate!
             if not property_renaming_config and predicate != RDF.type:
                 property_values[remove_namespace_from_uri(predicate, validation="prefix")].append(value)
-
+            elif predicate == RDF.type:
+                property_values[RDF.type].append(value)
```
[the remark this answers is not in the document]
doctrino: I think it is safer to keep this as is, as it the full URI of RDF.type cannot be used as a property. While the word 'InstanceType' can be a property of a View or any other random source.

PR #613, opened 2024-09-06
[NEAT-419] 🐳 Cleanup serialization Part 1

doctrino:
* Rename `SheetEntity` to `SheetRow` to more descriptive of what this is reprensenting.

* Rework `SheetList` this is now an actual Python list and thus gets all methods for free. Implemented custom `pydantic` schema such that it can be used with `pydantic`. This avoids the clumsy `data` property.

* Removed unused methods/functions I found in the process.

### DIFF
--- a/cognite/neat/rules/models/_base_rules.py
+++ b/cognite/neat/rules/models/_base_rules.py
@@ -4,12 +4,10 @@ its sub-models and validators.
 
 from __future__ import annotations
 
-import math
 import sys
 import types
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterator
-from functools import wraps
 from typing import Annotated, Any, ClassVar, Generic, Literal, TypeVar
 
 import pandas as pd
@@ -23,7 +21,6 @@ from pydantic import (
     model_serializer,
     model_validator,
 )
-from pydantic.fields import FieldInfo
 from pydantic.main import IncEx
 
 if sys.version_info >= (3, 11):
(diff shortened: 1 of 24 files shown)


cognite/neat/rules/models/_base_rules.py
```diff
@@ -148,7 +90,7 @@ class MatchType(StrEnum):
     partial = "partial"
 
 
-class RuleModel(BaseModel):
+class NeatModel(BaseModel):
```
nikokaoja, line 93: I am not 100% if this is a good name a bit confusing

cognite/neat/rules/models/_base_rules.py
```diff
@@ -281,46 +181,69 @@ def dump(
...
+            sheet_name = (field.alias or field_name) if by_alias else field_name
+            annotation = field.annotation
+
+            if isinstance(annotation, types.UnionType):
+                annotation = annotation.__args__[0]
+
+            try:
+                if isinstance(annotation, types.GenericAlias) and get_origin(annotation) is SheetList:
+                    # We know that this is a SheetList, so we can safely access the annotation
+                    # which is the concrete type of the SheetEntity.
```
nikokaoja, line 200: Isn't SheetEntity renamed to SheetRow?

cognite/neat/rules/models/_base_rules.py
```diff
@@ -281,46 +181,69 @@ def dump(
             exclude_defaults=exclude_defaults,
         )
 
+    @classmethod
+    def headers_by_sheet(cls, by_alias: bool = False) -> dict[str, list[str]]:
+        """Returns a list of headers for the model."""
```
nikokaoja, line 186: I need a bit of context where it is used
```suggestion
        """Returns a list of headers for the model, typically used by ExcelExporter"""
```

nikokaoja (approved):
To summarize,

It allows us access to `rows` directly such as rules.classes[0] instead of before rules.classes.data[0].

cognite/neat/rules/models/_base_rules.py
doctrino: Yes, it is. I think it is more descriptive of the source format

PR #616, opened 2024-09-08
[NEAT-424] 😎 Classic Importer

doctrino:
This is ready for review. It does not expose the new importer, so no addition to the changelog yet. 

It is not complete, that will be done later, after the end to end testing is done.


cognite/neat/constants.py
```diff
@@ -13,6 +13,7 @@
 
 DEFAULT_SPACE_URI = "http://purl.org/cognite/{space}#"
 DEFAULT_NAMESPACE = Namespace("http://purl.org/cognite/neat#")
+CLASSIC_CDF_NAMESPACE = Namespace("http://purl.org/cognite/cdf-classic#")
```
doctrino, line 16: Suggestion: Require this namespace in the ClassicImporter.

cognite/neat/graph/queries/_base.py
```diff
@@ -95,6 +95,42 @@ def triples_of_type_instances(self, rdf_type: str) -> list[tuple[str, str, str]]
             )
             return []
 
+    def types_with_property(self, property_uri: URIRef) -> list[URIRef]:
```
doctrino, line 98: These are a sew helpers.

cognite/neat/rules/importers/_cdf_classic.py
```diff
@@ -0,0 +1,231 @@
...
+        return ReadRules(rules, issues, read_context={})
+
+    def _add_source_system(self, rules: InformationInputRules) -> None:
+        types_ = self._store.queries.types_with_property(CLASSIC_CDF_NAMESPACE.source)
+        if not types_:
+            return
+        rules.classes.append(
+            InformationInputClass(
+                class_="SourceSystem",
+                parent="cdf_cdm:SourceSystem",
```
doctrino, line 48: Thinking is that we create classes and use the `parent` or `reference` to link it to the core model.

cognite/neat/rules/importers/_cdf_classic.py
```diff
@@ -0,0 +1,231 @@
...
+            ),
+        )
+        # TODO: Is this the correct syntax for the transformation?
+        transformation = [
+            f"classic_cdf:{remove_namespace_from_uri(type_)}(source) -> classic_cdf:SourceSystem(name)"
+            for type_ in types_
+        ]
```
doctrino, line 55: Then we use the transformation to describe how the mapping needs to be done.

For example, `Event`, `Asset`, and `FileMetadata` has ta property source which is a string. This should be mapped to SourceSystem in the new core model.

cognite/neat/rules/importers/_cdf_classic.py
```diff
@@ -0,0 +1,231 @@
...
+                parent=None,
+            ),
+        )
+        rules.properties.extend(
+            [
+                InformationInputProperty("DataSet", "name", "string"),
+                InformationInputProperty("DataSet", "description", "string"),
+                InformationInputProperty("DataSet", "external_id", "string"),
+                InformationInputProperty("DataSet", "write_protected", "boolean"),
+                InformationInputProperty("DataSet", "metadata", "json"),
```
doctrino, line 83: We can consider to check whether there is any instance that has a metadata property for the Asset type. Then only add this property if there are instances that have it.

cognite/neat/rules/importers/_cdf_classic.py
```diff
@@ -0,0 +1,231 @@
...
+                    "parent_id",
+                    "Asset",
+                    transformation="classic_cdf:Asset(parent_id) -> classic_cdf:Asset(parent)",
+                ),
+                InformationInputProperty("Asset", "description", "string"),
+                InformationInputProperty(
+                    "Asset",
+                    "dataSet",
+                    "DataSet",
+                    transformation="classic_cdf:Asset(dataSet_id) -> classic_cdf:Asset(dataSet)",
```
doctrino, line 115: @nikokaoja Need some help to check if I remember this syntax correctly.

cognite/neat/rules/importers/_cdf_classic.py
```diff
@@ -0,0 +1,231 @@
...
+                InformationInputProperty(
+                    "Asset",
+                    "tags",
+                    "string",
+                    max_count=10,
+                    transformation="classic_cdf:Asset(labels) -> classic_cdf:Asset(tags)",
+                ),
+            ]
+        )
+        # Todo Deal with GeoLocation
```
doctrino, line 128: We also need to look up all relationships starting with an Asset, and figure out how it should be added.

tests/data/classic_windfarm.py
```diff
@@ -0,0 +1,302 @@
+from cognite.client.data_classes import (
```
nikokaoja, line 1: As you have put some thoughts into this classic dataset I am wandering do we want to maintain `./tests/data/class_cdf_extractore_data`, which I created with a similar idea (when testing classic cdf extractors) but I stored instances/resources as YAML files ?

nikokaoja (approved):

PR #627, opened 2024-09-17
[NEAT-439] 🐛Strip cells

doctrino:
Multiple sources for the bug
1. In verified rules, only SheetRows were stripped not metadata sheets. Fixed by moving validator to appropriate location.
2. Input rules creates entities, and entities does not strip extra spaces. Fixed by stripping whitespace from entities.
3. Field validators in child classes have precedence over parent classes (new learning for me), so need to strip in the Enum conversion methods.

### DIFF
--- a/cognite/neat/rules/models/_base_rules.py
+++ b/cognite/neat/rules/models/_base_rules.py
@@ -108,6 +108,14 @@ class SchemaModel(BaseModel):
         """Returns a set of mandatory fields for the model."""
         return _get_required_fields(cls, use_alias)
 
+    @field_validator("*", mode="before")
+    def strip_string(cls, value: Any) -> Any:
+        if isinstance(value, str):
+            return value.strip()
+        elif isinstance(value, list):
+            return [entry.strip() if isinstance(entry, str) else entry for entry in value]
+        return value
+
 
 class BaseMetadata(SchemaModel):
     """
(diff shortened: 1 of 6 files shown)


cognite/neat/rules/models/_base_rules.py
```diff
@@ -108,6 +108,14 @@ def mandatory_fields(cls, use_alias=False) -> set[str]:
         """Returns a set of mandatory fields for the model."""
         return _get_required_fields(cls, use_alias)
 
+    @field_validator("*", mode="before")
```
doctrino, line 111: Fix one

cognite/neat/rules/models/entities/_single_value.py
```diff
@@ -97,6 +98,14 @@ def _load(cls, data: Any) -> "dict | Entity":
     def as_str(self) -> str:
         return str(self)
 
+    @field_validator("*", mode="before")
+    def strip_string(cls, value: Any) -> Any:
+        if isinstance(value, str):
+            return value.strip()
+        elif isinstance(value, list):
+            return [entry.strip() if isinstance(entry, str) else entry for entry in value]
+        return value
```
doctrino, line 107: Fix two

nikokaoja (approved):

PR #633, opened 2024-09-18
[NEAT-75] 👌End to end tests

doctrino:


tests/tests_end_to_end/test_graph_flow.py
```diff
@@ -0,0 +1,85 @@
...
+            cls_.class_ = new_name
+
+        # We need to filter out the externalID property from the rules
+        # as these are reserved in DMS.
+        new_properties: list[InformationInputProperty] = []
+        for prop in read_rules.rules.properties:
+            if prop.property_ not in RESERVED_PROPERTIES:
+                prop.class_ = naming_mapping[prop.class_]
+                new_properties.append(prop)
+        read_rules.rules.properties = new_properties
```
doctrino, line 61: Another issue we have to deal with when moving from Classic to DMS.

tests/tests_end_to_end/test_graph_flow.py
```diff
@@ -0,0 +1,85 @@
...
+        read_rules.rules.metadata.updated = "2024-09-19T00:00:00Z"
+
+        # We need to rename the classes to non-reserved names
+        naming_mapping: dict[str, str] = {}
+        for cls_ in read_rules.rules.classes:
+            new_name = f"Classic{cls_.class_}"
+            naming_mapping[cls_.class_] = new_name
+            cls_.class_ = new_name
+
+        # We need to filter out the externalID property from the rules
```
nikokaoja, line 54: do you maybe want to rename it ? prefixing property which are reserved with same way you are doing for classes ?

tests/tests_end_to_end/test_graph_flow.py
```diff
@@ -0,0 +1,85 @@
...
+
+from cognite.neat.graph.loaders import DMSLoader
+from cognite.neat.rules.exporters import YAMLExporter
+from cognite.neat.rules.importers import InferenceImporter
+from cognite.neat.rules.models.information import InformationInputProperty
+from cognite.neat.rules.transformers import InformationToDMS, VerifyInformationRules
+from cognite.neat.store import NeatGraphStore
+from tests.data import classic_windfarm
+
+RESERVED_PROPERTIES = frozenset(
```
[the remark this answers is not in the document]
doctrino: Suggest we keep it here for now, and fix this properly when we figure out how to handle the issue.

PR #634, opened 2024-09-19
Imf attribute types

nikokaoja:
- Interfaces for InferenceImporter and IMFImporter are leveraging BaseRDFImporter
- Renamed rules.examples to rules.catalog to start building catalog od data models
- Improved IMF rules that will handle IMF AttributeType onboarding


doctrino (approved):
Good to go

PR #637, opened 2024-09-20
Opinionated conversion of value type edge cases

nikokaoja:
Improved handling of unknown, multi-data, multi-objet and mixed value types in conversion from Information to DMS rules


doctrino (approved):
Elegant solution. 

Just some control questions (I think it makes sense what you have done). Is it correct to have the parameter `non_exisitng_node_type` in the RDF importers? Should it just be hardcoded, is there any reason for changing it? Does it belong in a RuleTransformer?

PR #644, opened 2024-10-03
[NEAT-425] 🗺Classic to Core Mapping Transformer

doctrino:
No longer a ClassicToCore mapper but a RuleMapper as it is completely generic


cognite/neat/rules/models/_base_rules.py
```diff
@@ -348,3 +350,13 @@ def __getitem__(self, index: SupportsIndex | slice, /) -> T_SheetRow | SheetList
...
+
+
+# Immutable such that this can be used as a key in a dictionary
+class PropertyRef(BaseModel, frozen=True):
+    class_: ClassEntityType = Field(alias="Class")
+    property_: InformationPropertyType = Field(alias="Property")
+
+
+class ClassRef(BaseModel, frozen=True):
+    class_: ClassEntityType = Field(alias="Class")
```
doctrino, line 362: I think these will be useful in more context. Main motivation for adding them to the `_base_rules` is to have the `as_reference()` methods in `InformationProperty` and `InformationClass`.

cognite/neat/rules/transformers/_mapping.py
```diff
@@ -95,3 +96,33 @@ def transform(self, rules: DMSRules | OutRules[DMSRules]) -> JustRules[DMSRules]
...
+
+    Args:
+        mapping: The mapping to use.
+
+    """
+
+    def __init__(self, mapping: RuleMapping) -> None:
+        self.mapping = mapping
+
+    def transform(self, rules: InformationRules | OutRules[InformationRules]) -> JustRules[InformationRules]:
```
doctrino, line 114: Switched this to operate on verified rules instead of input rules. The motivation is that since mapping is a pydantic class and already verified to ClassEntity, the code becomes a lot cleaner. Mapping is also something you probably do after you have validated, still up for discussion.

cognite/neat/rules/models/mapping.py
```diff
@@ -0,0 +1,67 @@
...
+
+T_Mapping = TypeVar("T_Mapping", bound=ClassRef | PropertyRef)
+
+
+class Mapping(BaseModel, Generic[T_Mapping]):
+    source: T_Mapping
+    destination: T_Mapping
+
+
+class MappingList(list, MutableSequence[Mapping[T_Mapping]]):
```
doctrino, line 18: I went for generic here. A bit more complex, but more reusable. 
Advantage: Easier to use (nice notebook output, useful helper method `as_destination_by_source`)
Con: Harder to instantiate.

I think we just make a bunch of factory method that does the instantiation easier for us.

cognite/neat/rules/transformers/_mapping.py
```diff
@@ -95,3 +97,37 @@ def transform(self, rules: DMSRules | OutRules[DMSRules]) -> JustRules[DMSRules]
...
+        input_rules = self._to_rules(rules)
+
+        destination_by_source = self.mapping.properties.as_destination_by_source()
+        destination_cls_by_source = self.mapping.classes.as_destination_by_source()
+        for prop in input_rules.properties:
+            if destination_prop := destination_by_source.get(prop.as_reference()):
+                prop.class_ = destination_prop.class_
+                prop.property_ = destination_prop.property_
```
nikokaoja, line 123: would at all be useful to add information to prop.comment ?

nikokaoja (approved):

cognite/neat/rules/transformers/_mapping.py
doctrino: Then we lose information that is already there. Or we start to clutter it. Suggest we think a bit carefully through how to do the lineage here.

PR #646, opened 2024-10-03
Neat 447 rules transformer dms compliant characters

nikokaoja:


cognite/neat/rules/transformers/_converters.py
```diff
@@ -61,7 +66,100 @@ def _transform(self, rules: T_VerifiedInRules) -> T_VerifiedOutRules:
         raise NotImplementedError()
 
 
-class ToCompliantEntityIds(ConversionTransformer[VerifiedRules, VerifiedRules]): ...
+class ToCompliantEntities(RulesTransformer[InformationInputRules, InformationInputRules]):  # type: ignore[misc]
+    """Converts input rules to rules with compliant entity IDs."""
```
doctrino, line 70: Is there some more documentation you can add here? Compliant with what?

Maybe just say which regex pattern you use with the properties

cognite/neat/rules/transformers/_converters.py
```diff
@@ -61,7 +66,100 @@ def _transform(self, rules: T_VerifiedInRules) -> T_VerifiedOutRules:
...
+                suffix=cls._fix_entity(class_.suffix),
+            )
+
+        return class_
+
+    @classmethod
+    def _fix_value_type(
+        cls, value_type: str | DataType | ClassEntity | MultiValueTypeInfo
+    ) -> str | DataType | ClassEntity | MultiValueTypeInfo:
+        fixed_value_type = value_type
```
doctrino, line 118: This is not used, you have a value type for each condition in the if-else below

cognite/neat/rules/_shared.py
```diff
@@ -18,6 +18,7 @@
 Rules: TypeAlias = (
     AssetInputRules | DMSInputRules | InformationInputRules | DomainRules | InformationRules | DMSRules | AssetRules
 )
+AllInformationRules: TypeAlias = InformationRules | InformationInputRules
```
doctrino, line 21: Looks like this is not used, so suggest removing it
```suggestion
```

cognite/neat/rules/transformers/_converters.py
nikokaoja: Will add good point

PR #647, opened 2024-10-04
Simplify RDF based importers

nikokaoja:
All RDF based importers will now automatically configure `transformation` column. Till now only Inference importer was doing this.


cognite/neat/rules/importers/_rdf/_base.py
```diff
@@ -104,7 +105,40 @@ def to_rules(
...
 
         rules = InformationInputRules.load(rules_dict)
+
+        self._add_transformation(rules)
+
         return ReadRules(rules, self.issue_list, {})
 
     def _to_rules_components(self) -> dict:
         raise NotImplementedError()
+
```
nikokaoja, line 115: this is now more explicit and easier way to build up both prefixes as well as transformations

doctrino (approved):

PR #649, opened 2024-10-04
Graph transformer multi value

nikokaoja:
Graph transformer `SplitMultiValueProperty` which splits multi-value properties into separate properties with single value

### DIFF
--- a/cognite/neat/graph/transformers/_value_type.py
+++ b/cognite/neat/graph/transformers/_value_type.py
@@ -2,6 +2,7 @@ from typing import cast
 
 from rdflib import XSD, Graph, URIRef
 
+from cognite.neat.issues._base import IssueList
 from cognite.neat.rules.analysis import InformationAnalysis
 from cognite.neat.rules.models._rdfpath import SingleProperty
 from cognite.neat.rules.models.data_types import AnyURI, DataType
@@ -38,32 +39,32 @@ class SplitMultiValueProperty(BaseTransformer):
 
                                 }}"""
 
-    def __init__(
-        self,
-        rules: InformationRules,
-    ):
-        self.rules = rules
-
     def transform(self, graph: Graph) -> None:
-        for property_ in InformationAnalysis(self.rules).multi_value_properties:
+        rules = self._infer_rules(graph)
+
+        class_reference_pairs = {class_.class_: class_.reference for class_ in rules.classes}
+
+        for property_ in InformationAnalysis(rules).multi_value_properties:
             transformation = property_.transformation
             if transformation and isinstance(transformation.traversal, SingleProperty):
                 traversal = transformation.traversal
 
                 for value_type in cast(MultiValueTypeInfo, property_.value_type).types:
-                    property_uri = self.rules.prefixes.get(traversal.property.prefix, self.rules.metadata.namespace)[
+                    property_uri = rules.prefixes.get(traversal.property.prefix, rules.metadata.namespace)[
                         traversal.property.suffix
                     ]
 
-                    subject_uri = self.rules.prefixes.get(traversal.class_.prefix, self.rules.metadata.namespace)[
+                    subject_uri = rules.prefixes.get(traversal.class_.prefix, rules.metadata.namespace)[
                         traversal.class_.suffix
                     ]
 
-                    if isinstance(value_type, ClassEntity):
-                        object_uri = self.rules.prefixes.get(
-                            cast(str, value_type.prefix), self.rules.metadata.namespace
-                        )[value_type.suffix]
-
+                    # needs forming object_uri this way since we are also
+                    # translating original rdf types to rdf types in new
+                    # namespace so we need to pick the correct one which are
+                    # store in class references
+                    if isinstance(value_type, ClassEntity) and (
+                        object_uri := class_reference_pairs.get(value_type, None)
+                    ):
                         for s, o in graph.query(  # type: ignore [misc]
                             self._object_property_template.format(
                                 subject_uri=subject_uri,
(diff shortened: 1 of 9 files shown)


tests/tests_unit/graph/test_transformers/test_split_multi_value_property_transformer.py
```diff
@@ -0,0 +1,18 @@
...
+
+def test_split_multi_value_property():
+    store = NeatGraphStore.from_oxi_store()
+    extractor = RdfFileExtractor(nordic44_knowledge_graph, base_uri="http://nordic44.com/")
+    store.write(extractor)
+
+    store.transform(SplitMultiValueProperty())
+
+    rules = ImporterPipeline.verify(InferenceImporter.from_graph_store(store))
+    assert len(InformationAnalysis(rules).multi_value_properties) == 0
```
doctrino, line 18: Nice test

cognite/neat/graph/transformers/_value_type.py
```diff
@@ -0,0 +1,110 @@
...
+from cognite.neat.rules.models.entities import (
+    ClassEntity,
+    MultiValueTypeInfo,
+)
+from cognite.neat.rules.models.information import InformationRules
+
+from ._base import BaseTransformer
+
+
+class SplitMultiValueProperty(BaseTransformer):
```
doctrino, line 18: Unrelated to this PR: What about renaming BaseTransformer to GraphTransformer to distinguish from RuleTransformer?

cognite/neat/graph/transformers/_value_type.py
```diff
@@ -0,0 +1,110 @@
...
+    _datatype_property_template: str = """SELECT ?s ?o WHERE {{
+
+                                ?s a <{subject_uri}> .
+                                ?s <{property_uri}> ?o .
+                                FILTER (datatype(?o) = <{datatype}>)
+
+                                }}"""
+
+    def transform(self, graph: Graph) -> None:
+        rules = self._infer_rules(graph)
```
doctrino, line 43: Isn't this an expensive operation to do? Isn't there simpler SPARQL queries to find multivalued properties?

I am afraid that for larger graph this will be very time consuming.

cognite/neat/graph/transformers/_value_type.py
```diff
@@ -0,0 +1,110 @@
...
+
+        class_reference_pairs = {class_.class_: class_.reference for class_ in rules.classes}
+
+        for property_ in InformationAnalysis(rules).multi_value_properties:
+            transformation = property_.transformation
+            if transformation and isinstance(transformation.traversal, SingleProperty):
```
doctrino, line 49: What is the reason for this check? Why is only SingleProperty traversal being splitted? I think there is an assumption here I am missing.

cognite/neat/graph/transformers/_value_type.py
```diff
@@ -0,0 +1,110 @@
...
+                traversal = transformation.traversal
+
+                for value_type in cast(MultiValueTypeInfo, property_.value_type).types:
+                    property_uri = rules.prefixes.get(traversal.property.prefix, rules.metadata.namespace)[
+                        traversal.property.suffix
+                    ]
+
+                    subject_uri = rules.prefixes.get(traversal.class_.prefix, rules.metadata.namespace)[
+                        traversal.class_.suffix
+                    ]
```
doctrino, line 59: These can be moved outside the loop.
```suggestion
                property_uri = rules.prefixes.get(traversal.property.prefix, rules.metadata.namespace)[
                    traversal.property.suffix
                ]

                subject_uri = rules.prefixes.get(traversal.class_.prefix, rules.metadata.namespace)[
                    traversal.class_.suffix
                ]
                for value_type in cast(MultiValueTypeInfo, property_.value_type).types:

```

cognite/neat/graph/transformers/_value_type.py
```diff
@@ -0,0 +1,110 @@
...
+
+                    # needs forming object_uri this way since we are also
+                    # translating original rdf types to rdf types in new
+                    # namespace so we need to pick the correct one which are
+                    # store in class references
+                    if isinstance(value_type, ClassEntity) and (
+                        object_uri := class_reference_pairs.get(value_type, None)
```
doctrino, line 66: None is the default so no need to specify

```suggestion
                        object_uri := class_reference_pairs.get(value_type)
```

cognite/neat/graph/transformers/_value_type.py
```diff
@@ -0,0 +1,110 @@
...
+                        for s, o in graph.query(  # type: ignore [misc]
+                            self._datatype_property_template.format(
+                                subject_uri=subject_uri,
+                                property_uri=property_uri,
+                                datatype=XSD[value_type.xsd],
+                            )
+                        ):
+                            graph.remove((s, property_uri, o))
+                            new_property = URIRef(f"{property_uri}_{value_type.xsd}")
+                            graph.add((s, new_property, o))
```
doctrino, line 89: Not sure if it is worth it, but warning of error there suddenly is something unexpected. 

The motivation for raising an error for a case that should not ever happen, is that in case it actually does you get a nice error message instead of something which is typically very hard to understand or you avoid silently failing. 

```suggestion
                            graph.add((s, new_property, o))
                    else:
                        raise ValueError(f"Value type {value_type} is not supported for splitting multi-value properties")
```

doctrino (commented):
There are a few things I need to understand on this one.

cognite/neat/graph/transformers/_value_type.py
nikokaoja: good point lets do that ! I will make maint. task in Jira board

cognite/neat/graph/transformers/_value_type.py
nikokaoja: outdated :)

cognite/neat/graph/transformers/_value_type.py
nikokaoja: outdated :)

doctrino (approved):

PR #654, opened 2024-10-11
[NEAT-467] 👫Relationships to Schema

doctrino:
Running into issues due to `RelationshipExtractor`: It does not link source_external_id and target_external_id as Entites, but instead uses literal for the external id.

### DIFF
--- a/cognite/neat/graph/queries/_base.py
+++ b/cognite/neat/graph/queries/_base.py
@@ -74,29 +74,32 @@ class Queries:
         # Select queries gives an iterable of result rows
         return cast(list[ResultRow], list(self.graph.query(query)))
 
-    def triples_of_type_instances(self, rdf_type: str) -> list[tuple[str, str, str]]:
+    def triples_of_type_instances(self, rdf_type: str | URIRef) -> list[tuple[str, str, str]]:
         """Get all triples of a given type.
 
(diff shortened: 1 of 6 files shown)


cognite/neat/graph/queries/_base.py
```diff
@@ -74,30 +74,33 @@ def list_instances_of_type(self, class_uri: URIRef) -> list[ResultRow]:
         # Select queries gives an iterable of result rows
         return cast(list[ResultRow], list(self.graph.query(query)))
 
-    def triples_of_type_instances(self, rdf_type: str) -> list[tuple[str, str, str]]:
+    def triples_of_type_instances(self, rdf_type: str | URIRef) -> list[tuple[str, str, str]]:
```
doctrino, line 77: Small extension of this method, used in experimentation. We can remove it, but I think it is an ok extension to include.

tests/data/classic_windfarm.py
```diff
@@ -161,7 +160,6 @@
     target_type="Asset",
     data_set_id=ds_source.id,
     labels=[Label("metMast")],
-    confidence=1.0,
```
doctrino, line 164: Removing these to avoid getting an edge

nikokaoja (approved):

PR #669, opened 2024-10-24
Session prepare

nikokaoja:
- added `prepare` endpoint
- added methods to push data model and instances to CDF
- Update docs to say that In Model is manadatory
- Fixed regex patterns 
- Addes specific types for data model external id and space (compliance with CDF API)

### DIFF
--- a/cognite/neat/_session/_state.py
+++ b/cognite/neat/_session/_state.py
@@ -39,13 +39,13 @@ class SessionState:
         return None
 
     @property
-    def verified_rule(self) -> VerifiedRules:
+    def last_verified_rule(self) -> VerifiedRules:
         if not self.verified_rules:
             raise ValueError("No verified rules provided")
         return self.verified_rules[-1]
 
     @property
-    def verifies_dms_rules(self) -> DMSRules | None:
+    def last_verified_dms_rules(self) -> DMSRules | None:
         if self.verified_rules:
             for rules in self.verified_rules[::-1]:
                 if isinstance(rules, DMSRules):
@@ -53,7 +53,7 @@ class SessionState:
         return None
 
     @property
-    def verifies_information_rules(self) -> InformationRules | None:
+    def last_verified_information_rules(self) -> InformationRules | None:
         if self.verified_rules:
             for rules in self.verified_rules[::-1]:
                 if isinstance(rules, InformationRules):
--- a/cognite/neat/_session/_state.py
+++ b/cognite/neat/_session/_state.py
@@ -2,6 +2,9 @@ from dataclasses import dataclass, field
 from typing import Literal, cast
 
 from cognite.neat._rules._shared import ReadRules, VerifiedRules
+from cognite.neat._rules.models.dms._rules import DMSRules
+from cognite.neat._rules.models.information._rules import InformationRules
+from cognite.neat._rules.models.information._rules_input import InformationInputRules
 from cognite.neat._store import NeatGraphStore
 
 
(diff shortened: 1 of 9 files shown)


cognite/neat/_session/_to.py
```diff
@@ -35,5 +37,30 @@ def yaml(self, io: Any | None = None) -> str | None:
...
+    def __init__(self, state: SessionState, client: CogniteClient | None, verbose: bool) -> None:
         self._client = client
+        self._state = state
+        self._verbose = verbose
+
+    def instances(self, space: str | None = None):
+        if not self._state.verifies_dms_rules:
+            raise ValueError("No verified DMS data model available")
+
+        space = space or self._state.verifies_dms_rules.metadata.space
```
doctrino, line 49: I think we should not do this. Instead lets prompt the user for instance space if they do not set it.

The `metadata.space` is a schema space and should not be used for data.

cognite/neat/_session/_to.py
```diff
@@ -35,5 +37,30 @@ def yaml(self, io: Any | None = None) -> str | None:
...
+        loader = loaders.DMSLoader.from_rules(self._state.verifies_dms_rules, self._state.store, space)
+
+        if self._client:
+            return loader.load_into_cdf(self._client)
+        else:
+            return loader.write_to_file(Path("./instance.yaml"))
```
doctrino, line 55: Isn't this confusing?  Shouldn't we instead have that placed somewhere else
```
neat.to.file.yaml(format="dms")
```

It is a bit unintuitive to me that if 
```
neat.to.cdf...()
```
I expect to push to CDF.

cognite/neat/_session/_to.py
```diff
@@ -35,5 +37,30 @@ def yaml(self, io: Any | None = None) -> str | None:
...
+
+    def data_model(self):
+        if not self._state.verifies_dms_rules:
+            raise ValueError("No verified DMS data model available")
+
+        exporter = DMSExporter()
+
+        if not self._client:
+            return exporter.export_to_file(self._state.verifies_dms_rules, Path("./data_model.zip"))
```
doctrino, line 64: Same here

cognite/neat/_session/_state.py
```diff
@@ -27,12 +30,36 @@ def input_rule(self) -> ReadRules:
...
+    @property
+    def verifies_dms_rules(self) -> DMSRules | None:
+        if self.verified_rules:
+            for rules in self.verified_rules[::-1]:
+                if isinstance(rules, DMSRules):
+                    return rules
+        return None
+
+    @property
+    def verifies_information_rules(self) -> InformationRules | None:
```
doctrino, line 56: Just to make it clear for us as developers.
```suggestion
    def last_verifies_information_rules(self) -> InformationRules | None:
```

cognite/neat/_session/_state.py
```diff
@@ -27,12 +30,36 @@ def input_rule(self) -> ReadRules:
...
+        return None
+
     @property
     def verified_rule(self) -> VerifiedRules:
         if not self.verified_rules:
             raise ValueError("No verified rules provided")
         return self.verified_rules[-1]
 
+    @property
+    def verifies_dms_rules(self) -> DMSRules | None:
```
doctrino, line 48:
```suggestion
    def last_verifies_dms_rules(self) -> DMSRules | None:
```

cognite/neat/_session/_state.py
```diff
@@ -27,12 +30,36 @@ def input_rule(self) -> ReadRules:
             raise ValueError("No input rules provided")
         return self.input_rules[-1]
 
+    @property
+    def information_input_rule(self) -> ReadRules | None:
```
doctrino, line 34:
```suggestion
    def last_information_input_rule(self) -> ReadRules | None:
```

doctrino (commented):
Good stuff. I think it is surprising that you suddenly write to file if you start with `neat.cdf.to`. Wdyt?

cognite/neat/_session/_to.py
nikokaoja: Agree !

cognite/neat/_session/_to.py
nikokaoja: Agree

doctrino (approved):

PR #671, opened 2024-10-25
Neat 486 add neat.rdf.examples.nordic44

nikokaoja:
- `NeatSession.read.rdf.examples.nordic44`
- Improved session overview in UI


doctrino (approved):
Nice improvement.

PR #672, opened 2024-10-25
[NEAT-463] Core Extension

doctrino:
image


cognite/neat/_rules/analysis/_dms.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @classmethod
+    def _set_cls_entity(cls, property_: DMSProperty, class_: ViewEntity) -> None:
+        property_.view = class_
+
+    def _get_object(self, property_: DMSProperty) -> ViewEntity | None:
+        return property_.value_type if isinstance(property_.value_type, ViewEntity) else None
+
+    def _get_max_occurrence(self, property_: DMSProperty) -> int | float | None:
+        return 1_000 if property_.is_list else 1
```
nikokaoja, line 39: is this actual limit for listable properties?

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -220,6 +224,140 @@ def _transform(self, rules: VerifiedRules) -> VerifiedRules:
         raise ValueError(f"Unsupported conversion from {type(rules)} to {self._out_cls}")
 
 
+_T_Entity = TypeVar("_T_Entity", bound=ClassEntity | ViewEntity)
+
+
+class ToExtension(RulesTransformer[DMSRules, DMSRules]):
+    def __init__(
+        self, new_model_id: DataModelIdentifier, prefix: str | None = None, mode: Literal["composition"] = "composition"
+    ):
```
nikokaoja, line 233: does prefix means the same thing like in case of Info Metadata?

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -220,6 +224,140 @@ def _transform(self, rules: VerifiedRules) -> VerifiedRules:
...
+        verified = self._to_rules(rules)
+        if self.prefix is None and verified.metadata.as_data_model_id() in COGNITE_MODELS:
+            raise NeatValueError(f"Prefix is required when extending {verified.metadata.as_data_model_id()}")
+        source_id = verified.metadata.as_data_model_id()
+
+        dump = verified.dump()
+        dump["metadata"]["schema_"] = SchemaCompleteness.partial.value
+        dump["metadata"]["space"] = self.new_model_id.space
+        dump["metadata"]["external_id"] = self.new_model_id.external_id
+        if self.new_model_id.version is not None:
```
nikokaoja, line 249: should we allow for new data model id without version ? I think we should force users to specify this

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -220,6 +224,140 @@ def _transform(self, rules: VerifiedRules) -> VerifiedRules:
...
+            dump["metadata"]["version"] = self.new_model_id.version
+        new_model = DMSRules.model_validate(DMSInputRules.load(dump).dump())
+
+        for prop in new_model.properties:
+            if prop.container and prop.container.space == self.new_model_id.space:
```
nikokaoja, line 254: I guess you need this since when you are dumping new data model id, container which do not have explicit statement of space will be automatically set to belong to new space, so you are reverting them back to the original config.

If so, just a comment above this for-loop would make code more readable

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -220,6 +224,140 @@ def _transform(self, rules: VerifiedRules) -> VerifiedRules:
...
+                prop.container = ContainerEntity(
+                    space=source_id.space,
+                    externalId=prop.container.suffix,
+                )
+
+        if self.mode == "composition":
```
nikokaoja, line 260: nice , since composition is nothing more than "copy paste"

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -220,6 +224,140 @@ def _transform(self, rules: VerifiedRules) -> VerifiedRules:
...
+            new_model.containers = None
+            for view in new_model.views:
+                view.implements = None
+
+        if source_id in COGNITE_MODELS:
```
nikokaoja, line 265: Nice, so if it is any of our data models we remove confusing prefix for end users

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -220,6 +224,140 @@ def _transform(self, rules: VerifiedRules) -> VerifiedRules:
...
+        self, new_model_id: DataModelIdentifier, prefix: str | None = None, mode: Literal["composition"] = "composition"
+    ):
+        self.new_model_id = DataModelId.load(new_model_id)
+        self.prefix = prefix
+        self.mode = mode
+
+    def transform(self, rules: DMSRules | OutRules[DMSRules]) -> JustRules[DMSRules]:
+        # Copy to ensure immutability
+        verified = self._to_rules(rules)
+        if self.prefix is None and verified.metadata.as_data_model_id() in COGNITE_MODELS:
```
nikokaoja, line 241: comment on the use case this covers would be helpful, as I can see below you allow for self.prefix to be None, but it is required if data model is actually Cognite_models. Why would it be, if you have option to strip `Cognite` from suffix if prefix is not provided. (maybe better name here is *affix* to make dist. from usage of prefix in Neat?)

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -220,6 +224,140 @@ def _transform(self, rules: VerifiedRules) -> VerifiedRules:
...
+                if isinstance(prop.value_type, ViewEntity):
+                    prop.value_type = self._remove_cognite_prefix(prop.value_type)
+            for view in new_model.views:
+                view.view = self._remove_cognite_prefix(view.view)
+                view.class_ = self._remove_cognite_prefix(view.class_)
+
+        return JustRules(new_model)
+
+    def _remove_cognite_prefix(self, entity: _T_Entity) -> _T_Entity:
+        new_suffix = entity.suffix.replace("Cognite", self.prefix or "")
```
nikokaoja, line 279: so if self.prefix is not provided you would remove `Cognite` , meaning L241 is not needed?

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -220,6 +224,140 @@ def _transform(self, rules: VerifiedRules) -> VerifiedRules:
...
+                ViewId("cdf_cdm", "CogniteDescribable", "v1"),
+                ViewId("cdf_cdm", "CogniteSchedulable", "v1"),
+                ViewId("cdf_cdm", "CogniteSourceable", "v1"),
+                ViewId("cdf_cdm", "CogniteVisualizable", "v1"),
+            }
+        ),
+    }
+
+    def __init__(self, drop: Collection[Literal["3D", "Annotation", "BaseViews"]]):
+        self.drop = drop
```
nikokaoja, line 327: nice control!

cognite/neat/_session/_prepare.py
```diff
@@ -31,3 +34,28 @@ def cdf_compliant_external_ids(self) -> None:
                     read_context={},
                 )
             )
+
+    def to_extension(self, new_data_model_id: DataModelIdentifier, prefix: str | None = None) -> None:
+        """Uses the current data model as a basis to extend from.
+
+        Args:
+            new_data_model_id: The new data model that is extending the current data model.
+            prefix: Prefix to use for the views in the new data model. This is required if the
```
nikokaoja, line 43: call it prefix here, but internally for us it is `affix` internally

nikokaoja (approved):
Great work!

See my comments, mainly readability of code base

cognite/neat/_rules/analysis/_dms.py
doctrino: yes, moving it out to make it clear

cognite/neat/_rules/transformers/_converters.py
doctrino: Renamed it org_name to make it clearer

PR #687, opened 2024-10-30
[NEAT-500] Update migration tutorial Part 1

doctrino:
Updated first migration tutorial to use `NeatSession`. A few changes

* Added `neat.read.cdf.class.asset()` to read assets from classic CDF.
* Added `neat.inspect.properties` to return a DataFrame with all properties.
* Made the visualization of the instances account for being less than 6 types.

### DIFF
--- a/cognite/neat/_session/_inspect.py
+++ b/cognite/neat/_session/_inspect.py
@@ -9,6 +9,7 @@ class InspectAPI:
     def __init__(self, state: SessionState) -> None:
         self._state = state
 
+    @property
     def properties(self) -> pd.DataFrame:
         """Returns the properties of the current data model."""
         return self._state.last_verified_rule.properties.to_pandas()
--- a/cognite/neat/_session/_inspect.py
+++ b/cognite/neat/_session/_inspect.py
@@ -1,0 +1,14 @@
+import pandas as pd
+
+from ._state import SessionState
+from .exceptions import intercept_session_exceptions
+
+
+@intercept_session_exceptions
+class InspectAPI:
+    def __init__(self, state: SessionState) -> None:
+        self._state = state
+
+    def properties(self) -> pd.DataFrame:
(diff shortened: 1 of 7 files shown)


cognite/neat/_session/_inspect.py
```diff
@@ -0,0 +1,15 @@
...
+
+@intercept_session_exceptions
+class InspectAPI:
+    def __init__(self, state: SessionState) -> None:
+        self._state = state
+
+    @property
+    def properties(self) -> pd.DataFrame:
+        """Returns the properties of the current data model."""
+        return self._state.last_verified_rule.properties.to_pandas()
```
nikokaoja, line 15: I've added Neat session error for last_verified rules so you have to check here if `self._state.last_verified_rule` does not raise error prior doing this

cognite/neat/_session/_read.py
```diff
@@ -52,16 +52,38 @@ def _return_filepath(self, io: Any) -> Path:
...
 class CDFReadAPI(BaseReadAPI):
-    def data_model(self, data_model_id: DataModelIdentifier) -> IssueList:
+    def __init__(self, state: SessionState, client: CogniteClient | None, verbose: bool) -> None:
+        super().__init__(state, client, verbose)
+        self.classic = CDFClassicAPI(state, client, verbose)
+
+    @property
+    def _get_client(self) -> CogniteClient:
         if self._client is None:
             raise ValueError("No client provided. Please provide a client to read a data model.")
```
nikokaoja, line 62: this should be NeatSessionError

nikokaoja (approved):
Approving but update code with NeatErrors/Warnings

cognite/neat/_session/_inspect.py
doctrino: I think we want to raise an error here if last_verified_rule is not there.

PR #689, opened 2024-10-30
Better handling of CDM extension

nikokaoja:


cognite/neat/_rules/transformers/_converters.py
```diff
@@ -232,56 +234,144 @@ def __init__(
...
+            view = DMSView(
+                view=view_entity,
+                implements=[definition.view],
+                in_model=True,
+                class_=class_entity,
+                name=definition.name,
+            )
 
-        return JustRules(new_model)
+            container = DMSContainer(
```
doctrino, line 342: Should an extension have containers? I don't think a pure extension should have containers at all, that is what you add when you actually do an extension

doctrino: I see this is connected to the property below. I don't understand this part.

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -232,56 +234,144 @@ def __init__(
...
+                container=container_entity,
+                class_=class_entity,
+            )
+
+            property_ = DMSProperty(
```
doctrino, line 347: What is the intention of adding a property per view to an extension model?

doctrino (commented):
I like the overall design, but a few open questions around extension.

cognite/neat/_rules/transformers/_converters.py
doctrino: Discussion with Fredrik now and I understand why you do this. Essentially a patch that bypasses the need for any fancy filters on the view. I see the value, going for this should have consequences how we set default filtering, as well as building solution models.

doctrino (approved):

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -232,56 +234,144 @@ def __init__(
         self,
         new_model_id: DataModelIdentifier,
         org_name: str | None = None,
-        mode: Literal["composition"] = "composition",
+        mode: Literal["extension", "solution"] = "extension",
     ):
         self.new_model_id = DataModelId.load(new_model_id)
+        if not self.new_model_id.version:
+            raise ValueError("Version is required for the new model.")
```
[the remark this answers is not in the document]
nikokaoja: yes! updated

cognite/neat/_session/_prepare.py
```diff
@@ -38,22 +38,68 @@ def cdf_compliant_external_ids(self) -> None:
...
     ) -> None:
-        """Uses the current data model as a basis to extend from.
+        """Uses the current data model as a basis to create enterprise data model
 
         Args:
-            new_data_model_id: The new data model that is extending the current data model.
-            org_name: Organization name to use for the views in the new data model. This is required if you are
-                creating an extension from a Cognite Data Model.
+            data_model_id: The enterprise data model id that is being created
+            org_name: Organization name to use for the views in the enterprise data model.
```
doctrino, line 51: Missing args in docs

cognite/neat/_session/_prepare.py
```diff
@@ -38,22 +38,68 @@ def cdf_compliant_external_ids(self) -> None:
...
+
+        !!! note "Enterprise Data Model Creation"
+            Always create an enterprise data model from a Cognite Data Model as this will
+            assure all the Cognite Data Fusion applications to run smoothly, such as
+            - Search
+            - Atlas AI
+            - ...
+
+        """
+        if dms := self._state.last_verified_dms_rules:
```
doctrino, line 61: I am a bit skeptical about this pattern we started to use as it will silently do nothing if there is no verified last dms rules.

cognite/neat/_session/_prepare.py
```diff
@@ -38,22 +38,68 @@ def cdf_compliant_external_ids(self) -> None:
...
+        org_name: str = "My",
+        mode: Literal["read", "write"] = "read",
+        dummy_property: str = "dummy",
+    ) -> None:
+        """Uses the current data model as a basis to create solution data model
+
+        Args:
+            data_model_id: The solution data model id that is being created.
+            org_name: Organization name to use for the views in the new data model.
+            mode: The mode of the solution data model. Can be either "read" or "write".
```
doctrino, line 82: dummy property missing. 

As well as docs on what it is

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -233,95 +234,167 @@ class ToExtension(RulesTransformer[DMSRules, DMSRules]):
...
-        mode: Literal["extension", "solution"] = "extension",
+        org_name: str = "My",
+        type_: Literal["enterprise", "solution"] = "enterprise",
+        mode: Literal["read", "write"] = "read",
+        dummy_property: str = "GUID",
     ):
         self.new_model_id = DataModelId.load(new_model_id)
         if not self.new_model_id.version:
-            raise ValueError("Version is required for the new model.")
+            raise NeatValueError("Version is required for the new model.")
```
doctrino, line 244: My personal preference is to only do validation in the transform method, not in the `__init__`.

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -233,95 +234,167 @@ class ToExtension(RulesTransformer[DMSRules, DMSRules]):
...
+        if self.type_ == "solution":
+            if self.mode not in ["read", "write"]:
+                raise NeatValueError(f"Unsupported mode: {self.mode}")
+
+            if reference_model_id in COGNITE_MODELS:
+                print(
+                    "You are building solution model on top of"
+                    " Cognite Data Model, this is not recommended."
+                    " Build the solution model on top of enterprise model instead."
+                )
```
doctrino, line 266: Should probably raise a warning here, and then do some magic in neat session to catch it.

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -233,95 +234,167 @@ class ToExtension(RulesTransformer[DMSRules, DMSRules]):
...
+        else:
+            raise NeatValueError(f"Unsupported data model type: {self.type_}")
 
-        elif self.mode == "extension":
-            return self._to_extension(new_model, reference_model_id)
+    def _has_views_in_multiple_space(self, rules: DMSRules) -> bool:
+        for view in rules.views:
+            if view.view.space != rules.metadata.space:
+                return True
+        return False
```
doctrino, line 283:
```suggestion
        return any(view.view.space != rules.metadata.space for view in rules.views)
```

doctrino (approved):

PR #693, opened 2024-10-31
[NEAT-487] Aggegated view of issues

doctrino:
image

### DIFF
--- a/cognite/neat/_constants.py
+++ b/cognite/neat/_constants.py
@@ -5,16 +5,21 @@ from rdflib import DCTERMS, OWL, RDF, RDFS, SKOS, XSD, Namespace, URIRef
 
 from cognite import neat
 
-IN_NOTEBOOK = True
-try:
-    from IPython import get_ipython
 
-    if "IPKernelApp" not in get_ipython().config:  # pragma: no cover
-        IN_NOTEBOOK = False
-except ImportError:
-    IN_NOTEBOOK = False
-except AttributeError:
-    IN_NOTEBOOK = False
+def _is_in_notebook() -> bool:
+    try:
+        from IPython import get_ipython
+
+        if "IPKernelApp" not in get_ipython().config:  # pragma: no cover
+            return False
+    except ImportError:
+        return False
+    except AttributeError:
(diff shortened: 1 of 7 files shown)


cognite/neat/_constants.py
```diff
@@ -5,6 +5,17 @@
 
 from cognite import neat
 
+IN_NOTEBOOK = True
```
nikokaoja, line 8: can you wrap it in small method ? that you call ?

cognite/neat/_session/_inspect.py
```diff
@@ -8,8 +14,76 @@
...
+    def __call__(
+        self,
+        search: str | None = None,
+        return_dataframe: Literal[True] = (False if IN_NOTEBOOK else True),  # type: ignore[assignment]
+    ) -> pd.DataFrame: ...
+
+    @overload
+    def __call__(
+        self,
+        search: str | None = None,
```
nikokaoja, line 42: nice feature,

cognite/neat/_session/_inspect.py
```diff
@@ -8,8 +14,76 @@
...
+        return_dataframe: bool = (False if IN_NOTEBOOK else True),  # type: ignore[assignment]
+    ) -> pd.DataFrame | None:
+        """Returns the issues of the current data model."""
+        issues = self._state.last_issues
+        if not issues:
+            self._print("No issues found.")
+
+        if issues and search is not None:
+            unique_types = {type(issue).__name__ for issue in issues}
+            closest_match = set(difflib.get_close_matches(search, unique_types))
```
nikokaoja, line 58: will this return partial matches ? what about fuzzy instead ? https://github.com/seatgeek/thefuzz or using cognite sdk entity matching ?

nikokaoja (approved):

cognite/neat/_session/_inspect.py
doctrino: This is a very simple way of using the standard library without any extra dependencies to get a slightly better search experience. Suggest we avoid any extra dependency

PR #712, opened 2024-11-05
Neat session rules provenance

nikokaoja:
image

We have following:

`Source Entity` which is used by `Activity` executed by `Agent` to generate `Target Entity`.

- Source Entity : Excel file, OWL, InputRules, ValidatedRules
- Agents: Rules transformers (e.g., verify, converters, reducers, etc.)
- Activity: conversion, verification, ...
- Target Entity: InputRules, ValidatedRules


doctrino (commented):
This is a very good work, very powerful concept to add to the code base.

There seems to be some inconsistencies with type. Should entity ids be string or URIRef. In addition, in the provenance method I see what I would consider invalid triples with `tuple[URIRef, URIRef, str | None]`. Is this intentional?

doctrino (approved):

PR #726, opened 2024-11-11
Validation of reverse connection feasibility in DMS Rules

nikokaoja:


cognite/neat/_rules/models/dms/_validation.py
```diff
@@ -334,6 +339,51 @@ def _validate_value_type_existence(self) -> None:
                     )
                 )
 
+    def _validate_reverse_connections(self) -> None:
+        # do not check for reverse connections in Cognite models
+        if self.metadata.as_data_model_id() in COGNITE_MODELS:
+            return None
+
+        properties_by_view = {f"{prop_.view.suffix}.{prop_.property_}": prop_ for prop_ in self.properties}
```
doctrino, line 347: This looks like properties by id not by view. Any reason not to use the full entity for the view? This is probably a safe assumption, but you are assuming unique external ID for the views, which would fail if you have two views from different spaces with the same ExternalID
```suggestion

        properties_by_view = {f"{prop_.view!s}.{prop_.property_}": prop_ for prop_ in self.properties}
```

cognite/neat/_rules/models/dms/_validation.py
```diff
@@ -334,6 +339,51 @@ def _validate_value_type_existence(self) -> None:
...
+        reversed_ids = {
+            id_
```
doctrino, line 349: Any reason not to store the reverse property here? since you are looking it up in the next step. 
```suggestion
            id_: prop_.connection

```
Then you don't have to do the casting in the loop below either as `reverse_ids: dict[str, ReverseConnectionEntity]`

doctrino (approved):
Good stuff. Some coding suggestion

PR #727, opened 2024-11-12
Neat 533 move connections from parent to child data model

nikokaoja:


cognite/neat/_rules/transformers/_converters.py
```diff
@@ -396,10 +398,17 @@ def _to_enterprise(self, reference_model: DMSRules) -> JustRules[DMSRules]:
         # extending reference views with new ones
         enterprise_model.views.extend(enterprise_views)
 
+        # Move connections from reference model to enterprise model
+        if self.move_connections:
+            enterprise_connections = self._move_connections(enterprise_model)
```
doctrino, line 403: I guess `mypy` is not too happy about this.
```suggestion
        if self.move_connections:
            enterprise_connections = self._move_connections(enterprise_model)
        else:
            enterprise_connections = SheetList[DMSProperty]()
```

doctrino (approved):

PR #732, opened 2024-11-13
[NEAT-536] 🧑‍🏭Download from GitHub

doctrino:
image


nikokaoja (approved):
Make sure when one passes github link make sure to check that the form of URL is

https://github.com/org/repo/blob/branch/.../somefile

as you might have ppl actually pasting 

https://raw.github....

PR #745, opened 2024-11-14
add xml support

Juliamg:


doctrino (commented):
Lets include a changelog entry

docs/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Changes are grouped as follows:
 ## TBD
 ### Fixed
 - `neat.set.data_model_id()` did not result in the verified data model being stored
+- Support for XML/Dexpi files so that you can now run `neat.read.xml()`. Only supports DEXPI atm.
```
Juliamg, line 21: like so? @doctrino

cognite/neat/_session/_read.py
```diff
@@ -167,6 +168,27 @@ def __call__(self, io: Any, type: str, primary_key: str) -> None:
         self._state.instances.store.write(extractor)
 
 
+@intercept_session_exceptions
+class XMLReadAPI(BaseReadAPI):
```
nikokaoja, line 172: Suggestion also implement it similar to what we did for RDF:

https://github.com/cognitedata/neat/blob/51a863bce49e014ebb7d1f1a7339d36a74abe528/cognite/neat/_session/_read.py#L218C1-L244C84

nikokaoja (changes requested):
Suggestion same interface as for read.rdf

cognite/neat/_session/_read.py
```diff
@@ -167,6 +168,35 @@ def __call__(self, io: Any, type: str, primary_key: str) -> None:
...
+        if format.lower() != "dexpi":
+            raise NeatValueError("Only support XML files of DEXPI format at the moment.")
+
+        engine = import_engine()
+        engine.set.source = format
+        engine.set.file = path
+        extractor = engine.create_extractor()
+
+        self._state.instances.store.write(extractor)
+
```
nikokaoja, line 198:
```suggestion
        if format.lower() == "dexpi":
            return self.dexpi(path)
         else:
            raise NeatValueError("Only support XML files of DEXPI format at the moment.")
       
    def dexpi(self, path):  
     
        engine = import_engine()
        engine.set.source = "dexpi"
        engine.set.file = path
        extractor = engine.create_extractor()
        self._state.instances.store.write(extractor)
```

nikokaoja (approved):

PR #760, opened 2024-11-18
Make metadata same accross roles

nikokaoja:
- removed Domain rules
- removed Asset rules
- removed Asset loader
- replace prefix with space in information metadata
- added external_id to Information metadata
- `namespace` now automatically generated not anymore settable
- disable conversion of "reference" between roles since reference is over-used
- disable number of tests until we refactor rules
- rules aspect added as field to metadata
- linking of rules between roles enabled through optional metadata fields (.physical, .logical, .conceptual)


tests/tests_end_to_end/test_graph_flow.py
```diff
@@ -37,6 +38,7 @@
 )
 
 
+@pytest.mark.skip(reason="Skipping this test until refactoring of rules is completed")
```
doctrino, line 41: Does this mean that we are in a non-releaseable state?

docs/artifacts/rules/~$information-addition-svein-harald.xlsx
doctrino, line 1: Temporary file. Should not be committed. Suggest adding something like `~*.xlsx` to `.gitignore`.

cognite/neat/_rules/models/_base_rules.py
```diff
@@ -125,6 +150,34 @@ class BaseMetadata(SchemaModel):
     """
 
     role: ClassVar[RoleTypes]
+    aspect: ClassVar[DataModelAspect]
+    space: SpaceType = Field(SchemaCompleteness.partial, alias="prefix")
```
doctrino, line 154: Wrong default value?

cognite/neat/_rules/models/_base_rules.py
```diff
@@ -143,15 +196,30 @@ def mandatory_fields(cls, use_alias=False) -> set[str]:
...
+    def prefix(self) -> str:
+        return self.space
+
     def as_identifier(self) -> str:
-        """Returns a unique identifier for the metadata."""
-        raise NotImplementedError()
+        return f"{self.prefix}:{self.external_id}"
 
-    @abstractmethod
     def get_prefix(self) -> str:
```
doctrino, line 206: Lets remove this and replace it with .prefix. It is only used in one location

cognite/neat/_rules/models/_base_rules.py
```diff
@@ -143,15 +196,30 @@ def mandatory_fields(cls, use_alias=False) -> set[str]:
...
+
+    @property
+    def identifier(self) -> URIRef:
+        """Globally unique identifier for the data model.
+
+        !!! note
+            Unlike namespace, the identifier does not end with "/" or "#".
+
+        """
+        return DEFAULT_NAMESPACE[f"data-model/verified/{self.aspect}/{self.space}/{self.external_id}/{self.version}"]
```
doctrino, line 217: I like this! Very good idea to have aspect as part of the unique identifier.

cognite/neat/_rules/models/information/_rules_input.py
```diff
@@ -36,8 +36,8 @@ class InformationInputMetadata(InputComponent[InformationMetadata]):
     description: str | None = None
     created: datetime | str | None = None
     updated: datetime | str | None = None
-    license: str | None = None
-    rights: str | None = None
+    physical: str | None = None
+    conceptual: str | None = None
```
doctrino, line 40: Is the thinking here to use these two fields to link the different input rules?

cognite/neat/_rules/importers/_rdf/_shared.py
```diff
@@ -401,25 +376,12 @@ def fix_description(metadata: dict, default: str = "This model has been inferred
...
-def fix_namespace(metadata: dict, default: Namespace) -> dict:
-    if namespace := metadata.get("namespace", None):
-        if not isinstance(namespace, Namespace):
-            try:
-                metadata["namespace"] = Namespace(namespace)
-            except Exception:
-                metadata["namespace"] = default
+def fix_space(metadata: dict, default: str = "neat") -> dict:
+    if space := metadata.get("space", None):
+        if not isinstance(space, str) or not PATTERNS.space_compliance.match(space):
```
doctrino, line 381: A consequence now is that Information rules needs to use a prefix that is compliant with the regex for CDF space?

doctrino (commented):
Mostly clarification questions

doctrino (approved):

cognite/neat/_rules/models/_base_rules.py
nikokaoja: Good catch !

cognite/neat/_rules/models/information/_rules_input.py
nikokaoja: Correct !

cognite/neat/_rules/importers/_rdf/_shared.py
nikokaoja: Yes, we are forcing users to closed world

PR #766, opened 2024-11-21
[NEAT-557] 😵Drop instances

doctrino:
image


cognite/neat/_graph/queries/_base.py
```diff
@@ -342,3 +342,20 @@ def multi_value_type_property(
...
+
+        Returns:
+            Dictionary of dropped types
+        """
+        dropped_types: dict[URIRef, int] = {}
+        for t in type_:
+            triples = self.list_instances_of_type(t)
+            dropped_types[t] = sum(1 for triple in triples if triple[1] is RDF.type)
+            for triple in triples:
+                self.graph.remove(triple)  # type: ignore[arg-type]
```
nikokaoja, line 360: This will only remove triple that specify type instead of all triples

```suggestion
    instances = self.queries.list_instances_ids_of_class(t)
    for instance in instances:
        store.graph.remove((instance, RDF.type, t))
```
Screenshot 2024-11-21 at 12 13 10

nikokaoja: Would be good to add this method next to `add_triples_in_batch` to get speed increase

```
def remove_triples_in_batch(graph: Graph, triples: Iterable[Triple], batch_size: int = 10_000):
    """Remove triples from the graph store in batches.

    Args:
        triples: list of triples to be removed to the graph store
        batch_size: Batch size of triples per commit, by default 10_000
        verbose: Verbose mode, by default False
    """

    commit_counter = 0
    number_of_written_triples = 0

    def check_commit(force_commit: bool = False):
        """Commit nodes to the graph if batch counter is reached or if force_commit is True"""
        nonlocal commit_counter
        nonlocal number_of_written_triples
        if force_commit:
            number_of_written_triples += commit_counter
            graph.commit()
            return
        commit_counter += 1
        if commit_counter >= batch_size:
            number_of_written_triples += commit_counter
            graph.commit()
            commit_counter = 0

    for triple in triples:
        graph.remove(triple)
        check_commit()

    check_commit(force_commit=True)
```

nikokaoja (commented):
see my comments below

nikokaoja (approved):

PR #767, opened 2024-11-21
[NEAT-558] 🗺 Prepare Classic to Core Mapping

doctrino:


cognite/neat/_session/_mapping.py
```diff
@@ -0,0 +1,21 @@
+from cognite.neat._rules.models.mapping import create_classic_to_core_mapping
+from cognite.neat._rules.transformers import RuleMapper
+
+from ._state import SessionState
+from .exceptions import session_class_wrapper
+
+
+@session_class_wrapper
+class MappingAPI:
```
doctrino, line 9: This is not connected to anything. So save to merge

cognite/neat/_rules/_constants.py
```diff
@@ -42,7 +43,7 @@ class EntityTypes(StrEnum):
     space = "space"
 
 
-def get_reserved_words() -> dict[str, list[str]]:
+def get_reserved_words(key: Literal["class", "view", "property", "space"]) -> list[str]:
```
doctrino, line 46: This was a change I did in preparation for having a mode on the prefix method to only prefix reserved keywords. Decided that it an unnecessary complexity, but the change is a simplification so think we should keep it.

nikokaoja (approved):

PR #772, opened 2024-11-22
[NEAT-558] 🗺 Classic to Core Mapping

doctrino:


cognite/neat/_rules/models/information/_rules.py
```diff
@@ -103,9 +101,6 @@ def remove_default_prefixes(self, value: Any, info: SerializationInfo) -> str:
             )
         return ",".join(str(value) for value in value)
 
-    def as_reference(self) -> ClassRef:
```
nikokaoja, line 106: Is there a reason why we are removing this? I guess we are not using it or ?

cognite/neat/_rules/models/information/_rules.py
```diff
@@ -233,9 +228,6 @@ def is_list(self) -> bool:
             isinstance(self.max_count, int | float) and self.max_count > 1
         )
 
-    def as_reference(self) -> PropertyRef:
```
nikokaoja, line 236: Is there a reason why we are removing this? I guess we are not using it or ?

scripts/create_classic_core_mapping_yaml.py
```diff
@@ -0,0 +1,29 @@
+from cognite.neat import NeatSession, get_cognite_client
+from pathlib import Path
+from rich import print
+THIS_FOLDER = Path(__file__).resolve().parent
+
+XLSX_FILE = THIS_FOLDER / "core_classic_mapping.xlsx"
```
nikokaoja, line 6: Not sure 100% how XLSX and YAML differ

nikokaoja (approved):

cognite/neat/_rules/models/information/_rules.py
doctrino: It is connected to the Mapping class which I removed since we are now using DMS rules for mapping instaed.

scripts/create_classic_core_mapping_yaml.py
doctrino: They don't. Xlsx is better for to work in, but is slow to load so do not want to include an xlsx file in the core of the code base.

PR #775, opened 2024-11-22
[NEAT-521] 🤔Fix view filter

doctrino:
The issue of node filter warnings was already solved, but if you set a filter it was ignored. This fixes that. 
In addition, skips creating node types for all views as this is no longer necessary.

### DIFF
--- a/cognite/neat/_rules/models/dms/_exporter.py
+++ b/cognite/neat/_rules/models/dms/_exporter.py
@@ -79,9 +79,11 @@ class _DMSExporter:
             view_properties_by_id, rules.views
         )
 
-        views, view_node_type_filters = self._create_views_with_node_types(
-            view_properties_by_id, view_properties_with_ancestors_by_id
-        )
+        views = self._create_views_with_node_types(view_properties_by_id, view_properties_with_ancestors_by_id)
+        view_node_type_filters: set[dm.NodeId] = set()
+        for dms_view in rules.views:
+            if isinstance(dms_view.filter_, NodeTypeFilter):
+                view_node_type_filters.update(node.as_id() for node in dms_view.filter_.inner or [])
         if rules.nodes:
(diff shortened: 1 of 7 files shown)


cognite/neat/_rules/models/dms/_exporter.py
```diff
@@ -155,11 +157,7 @@ def _create_views_with_node_types(
                 if view_property is not None:
                     view.properties[prop.view_property] = view_property
 
-        unique_node_types: set[dm.NodeId] = set()
-        for view in views.values():
-            unique_node_types.add(dm.NodeId(space=view.space, external_id=view.external_id))
```
doctrino, line 160: No longer necessary

cognite/neat/_session/_state.py
```diff
@@ -162,10 +161,10 @@ def lookup_containers(
         if missing := set(container_ids) - set(self._cdf_containers.keys()):
             cdf_container_ids = list(missing)
             found = client.data_modeling.containers.retrieve(cdf_container_ids).as_write()
-            if not_found := missing - set(found.as_ids()):
```
doctrino, line 165: Realized that a created a bug in the last PR. If a container or view is not in CDF (or cannot be accessed), we should let the regular post validation fail

cognite/neat/_rules/models/dms/_rules.py
```diff
@@ -272,6 +272,7 @@ def as_view(self) -> dm.ViewApply:
             version=view_id.version or _DEFAULT_VERSION,
             name=self.name or None,
             description=self.description,
+            filter=None if self.filter_ is None else self.filter_.as_dms_filter(),
```
doctrino, line 275: This was missing. I think it is ok to still support HasData and NodeFilter.

nikokaoja (approved):

PR #791, opened 2024-11-29
🧠 Improve inference

nikokaoja:
### Improved
- Inference, instances are ordered by no of properties
- Inference, if Unknown value is present in multivalue property it is dropped

### DIFF
--- a/cognite/neat/_rules/importers/_rdf/_inference2rules.py
+++ b/cognite/neat/_rules/importers/_rdf/_inference2rules.py
@@ -164,8 +164,8 @@ class InferenceImporter(BaseRDFImporter):
         # Infers all the properties of the class
         for class_id, class_definition in classes.items():
             for (  # type: ignore[misc]
-                instance,  # type: ignore[misc]
-                _,  # type: ignore[misc]
+                instance,
+                _,
             ) in self.graph.query(  # type: ignore[misc]
                 INSTANCES_OF_CLASS_QUERY.replace("class", class_definition["uri"])
                 if self.max_number_of_instance < 0
--- a/cognite/neat/_rules/importers/_rdf/_inference2rules.py
+++ b/cognite/neat/_rules/importers/_rdf/_inference2rules.py
@@ -27,10 +27,13 @@ ORDERED_CLASSES_QUERY = """SELECT ?class (count(?s) as ?instances )
                            group by ?class order by DESC(?instances)"""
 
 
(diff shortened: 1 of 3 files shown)


cognite/neat/_rules/importers/_rdf/_inference2rules.py
```diff
@@ -26,7 +26,11 @@
                            WHERE { ?s a ?class . }
                            group by ?class order by DESC(?instances)"""
 
-INSTANCES_OF_CLASS_QUERY = """SELECT ?s WHERE { ?s a <class> . }"""
+
+INSTANCES_OF_CLASS_QUERY = """SELECT ?s (COUNT(?p) as ?propertyCount)
```
nikokaoja, line 30: main 🧠 , a bit slower query than the previous, but here you can really count on getting the most richest instances first.

tests/tests_unit/rules/test_importers/test_inference_importer.py
```diff
@@ -46,6 +48,28 @@ def test_rdf_inference():
     assert len(InformationAnalysis(rules).multi_value_properties) == 4
 
 
+def test_rdf_inference_with_removal_of_unknown_type():
+    EX = Namespace("http://example.org/")
+    SUBSTATION = Namespace("http://example.org/substation/")
+    TERMINAL = Namespace("http://example.org/terminal/")
+    store = NeatGraphStore.from_oxi_store()
+
```
nikokaoja, line 56: you can see that here one of substations points to non-existing terminal, but resulting data model does not have multi-value type since Uknown type is dropped

cognite/neat/_rules/importers/_rdf/_inference2rules.py
```diff
@@ -242,6 +249,10 @@ def _to_rules_components(
 
         # Create multi-value properties otherwise single value
         for property_ in properties.values():
+            # Removes non-existing node type from value type prior final conversion to string
+            if len(property_["value_type"]) > 1 and str(self.non_existing_node_type) in property_["value_type"]:
```
nikokaoja, line 253: dropping unknown value type if value type is multi-value

doctrino (approved):
code base is good when you can do such a nice surgical change. Good stuff!

cognite/neat/_rules/importers/_rdf/_inference2rules.py
nikokaoja: SMART !!!  🧠 🧠 🧠 🧠 🧠

PR #821, opened 2024-12-07
[NEAT-566] 💊 Literal conversion

doctrino:
## In notebook on large example
image

## Before
image

## After
image


cognite/neat/_session/_collector.py
```diff
@@ -60,7 +60,7 @@ def track_session_command(self, command: str, *args, **kwargs) -> None:
         if len(args) > 1:
             # The first argument is self.
             for i, arg in enumerate(args[1:]):
-                event_information[f"arg{i}"] = arg
+                event_information[f"arg{i}"] = self._serialize_value(arg)[:500]
```
doctrino, line 63: Bug forgot to serialize on positional args.

cognite/neat/_session/_collector.py
```diff
@@ -73,6 +73,8 @@ def _serialize_value(value: Any) -> str:
             return str(value)
         if isinstance(value, list | tuple | dict):
             return str(value)
+        if callable(value):
+            return value.__name__
```
doctrino, line 77: When sending to mixpanel, we use the name of the callable.

cognite/neat/_session/_prepare.py
```diff
@@ -133,6 +132,32 @@ def relationships_as_edges(self, min_relationship_types: int = 1, limit_per_type
         transformer = RelationshipAsEdgeTransformer(min_relationship_types, limit_per_type)
         self._state.instances.store.transform(transformer)
 
+    def convert_data_type(self, source: tuple[str, str], *, convert: Callable[[Any], Any]) -> None:
```
doctrino, line 135: Note that the callable here is a deviation from requiring only primitively types as input in session method.

tests/data/classic_windfarm.py
```diff
@@ -218,6 +218,7 @@
 ]
 
 wind_turbine_production = TimeSeries(
+    id=1,
```
doctrino, line 221: Bug in test data missing identifier used in the extractor.

tests/tests_integration/test_session/test_graph_flow/test_classic_to_dms.yml
```diff
@@ -403,8 +403,11 @@ rules:
     value_type: ClassicAsset(version=v1_prefixed_Classic)
     view: ClassicTimeSeries(version=v1_prefixed_Classic)
     view_property: assets
-  - connection: edge
-    is_list: true
+  - connection: direct
+    container: ClassicTimeSeries
+    container_property: dataSetId
+    is_list: false
+    nullable: true
```
doctrino, line 410: After fixing the bug in the test data this changed

cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -64,3 +71,73 @@ def transform(self, graph: Graph) -> None:
...
+    WHERE {{
+      ?instance a <{subject_type}> .
+      ?instance <{subject_predicate}> ?property
+    }}"""
+
+    _properties = """SELECT ?instance ?property
+    WHERE {{
+      ?instance a <{subject_type}> .
+      ?instance <{subject_predicate}> ?property
+    }}"""
```
nikokaoja, line 91: This will now only return you values which are literals
```suggestion
    _properties = """SELECT ?instance ?value
    WHERE {{
      ?instance a <{subject_type}> .
      ?instance <{subject_predicate}> ?value
      
      FILTER(isIRI(?value))
      
    }}"""
```

cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -64,3 +71,73 @@ def transform(self, graph: Graph) -> None:
...
+class ConvertLiteral(BaseTransformer):
+    description: str = "ConvertLiteral is a transformer that converts a literal value."
+    _use_only_once: bool = False
+    _need_changes = frozenset({})
+
+    _count_by_properties = """SELECT (COUNT(?property) AS ?propertyCount)
+    WHERE {{
+      ?instance a <{subject_type}> .
+      ?instance <{subject_predicate}> ?property
+    }}"""
```
nikokaoja, line 85:
```suggestion
    _count_by_properties = """SELECT (COUNT(?value) AS ?valueCount)
    WHERE {{
      ?instance a <{subject_type}> .
      ?instance <{subject_predicate}> ?value
       FILTER(isIRI(?value))
    }}"""
```

cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -64,3 +71,73 @@ def transform(self, graph: Graph) -> None:
                     graph.remove((s, property_uri, o))
                     new_property = URIRef(f"{property_uri}_{remove_namespace_from_uri(value_type_uri)}")
                     graph.add((s, new_property, o))
+
+
+class ConvertLiteral(BaseTransformer):
+    description: str = "ConvertLiteral is a transformer that converts a literal value."
```
nikokaoja, line 77: This convert essentially is data-typing literal values, probably converting strings to the best data type possible (e.g., bool, float, etc.). Therefore name of the transformer should highlight this.

Maybe: DatatypeLiterals 

```suggestion
    description: str = "ConvertLiteral is a transformer that improve data typing of a literal value."
```

cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -64,3 +71,73 @@ def transform(self, graph: Graph) -> None:
...
+        iterate_query = self._properties.format(
+            subject_type=self.subject_type, subject_predicate=self.subject_predicate
+        )
+
+        for instance, literal in iterate_progress_bar(  # type: ignore[misc]
+            graph.query(iterate_query),
+            total=property_count,
+            description=f"Converting {self._type_name}.{self._property_name}.",
+        ):
+            if not isinstance(literal, rdflib.Literal):
```
nikokaoja, line 121: you do not need this if when you introduce:
```
...
      FILTER(isIRI(?value))
```

cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -64,3 +71,73 @@ def transform(self, graph: Graph) -> None:
                     graph.remove((s, property_uri, o))
                     new_property = URIRef(f"{property_uri}_{remove_namespace_from_uri(value_type_uri)}")
                     graph.add((s, new_property, o))
+
+
```
nikokaoja, line 75: This is useful transformer when one is unable to attach datatype to values when extracting triples from source

cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -64,3 +71,73 @@ def transform(self, graph: Graph) -> None:
...
+        self,
+        subject_type: URIRef,
+        subject_predicate: URIRef,
+        conversion: Callable[[Any], Any],
+    ) -> None:
+        self.subject_type = subject_type
+        self.subject_predicate = subject_predicate
+        self.conversion = conversion
+        self._type_name = remove_namespace_from_uri(subject_type)
+        self._property_name = remove_namespace_from_uri(subject_predicate)
```
nikokaoja, line 103: Since you will use filer no need for these
```suggestion
```

cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -64,3 +71,73 @@ def transform(self, graph: Graph) -> None:
...
+
+    def __init__(
+        self,
+        subject_type: URIRef,
+        subject_predicate: URIRef,
+        conversion: Callable[[Any], Any],
+    ) -> None:
+        self.subject_type = subject_type
+        self.subject_predicate = subject_predicate
+        self.conversion = conversion
```
nikokaoja, line 101: you should default conversion to `string_to_ideal_type` which is in `cognite.neat._utils.auxliliar`

cognite/neat/_session/_prepare.py
nikokaoja: see my comment on defaulting to conversion method

nikokaoja (changes requested):

cognite/neat/_graph/transformers/_value_type.py
doctrino: That is a nice learning. A few things, first shouldn't it be `Filter(isLiteral(?value))`. 

Second, is the user select a property which is a connection, this filter will silently ignore it which can lead to confusion. I think the current solution has the weakness that it will spam warnings for every case which is wrong. I suggest that I do a check first to see if there are any URI for the property, and then issue a warning for that? Wdyt?

cognite/neat/_graph/transformers/_value_type.py
doctrino: As I understand it, this will not solve the motivating use case for this transformer. In this case, we are importing from classic TimeSeries which has isString as a boolean. While, in CogniteTimeSeries we need this to be an enum with value 'string' and 'numeric'.

nikokaoja (approved):

PR #822, opened 2024-12-07
🍾 Extend capabilities of AttachPropertyFromTargetToSource

nikokaoja:


cognite/neat/_graph/transformers/_prune_graph.py
```diff
@@ -55,12 +56,13 @@ class AttachPropertyFromTargetToSource(BaseTransformer):
 
     Args:
         target_node_type: RDF.type of edge Node
-        namespace: RDF Namespace to use when querying the graph
-        target_property: str with name of the property that holds the value attached to the intermediate node
-        target_property_holding_new_property_name: Optional str of the property name that holds
-        the new predicate to use when attaching the new property to the source node.
+        target_property: URIRef of the property that holds the value attached to the intermediate node
+        target_property_holding_new_property_name: URIRef of the property which value will be new
```
doctrino, line 60:
```suggestion
        target_property_holding_new_property: URIRef of the property which value will be new

```

tests/tests_unit/graph/test_transformers/test_prune_graph.py
```diff
@@ -185,22 +340,24 @@ def test_two_hop_flattener_keep_connecting_node(
         target_node_type: URIRef,
         namespace: Namespace,
         target_property: str,
-        target_property_holding_new_property_name: str | None,
+        target_property_holding_new_property: str | None,
         expected_triples: list[Triple],
+        convert_literal_to_uri: bool,
```
doctrino, line 345: Not necessary, but I think these tests can be further simplified by creating the `AttachPropertyFromTargetToSource` transformer in the test cases, and pass the entire object instead of all the individual parameters. Typically, you want to avoid doing expensive operations in the construction of the test cases, but since there is no logic in the `__init__` of `AttachPropertyFromTargetToSource` this is not a problem.

```python
def test_two_hop_flattener_keep_connecting_node(
        self,
        triples: list[Triple],
        transformer: AttachPropertyFromTargetToSource,
        expected_triples: list[Triple],
    ) -> None:
 ....

doctrino (approved):
A few suggestions.

tests/tests_unit/graph/test_transformers/test_prune_graph.py
nikokaoja: Agree , and there are a few more improvements such as better bundling of initial triples and expected.

PR #823, opened 2024-12-07
[NEAT-623] 🤡 DMS Loader container constraints

doctrino:
Most of this PR is update of test data.

Two changes to DMSLoader
1. Only views with instances are run
2. Views are run in order of dependencies between views.

image


Juliamg (approved):
lgtm

PR #836, opened 2024-12-12
Reorg Docs

doctrino:
image


Juliamg (approved):
Makes sense, and I like the new titles

PR #837, opened 2024-12-12
[NEAT-639] 🥶 Data modeling cold start tutorial

doctrino:
image


cognite/neat/_client/_api/data_modeling_loaders.py
```diff
@@ -609,6 +609,7 @@ def _retrieve_recursive(
                 elif view_id in self._items_by_id:
                     view = self._items_by_id[view_id]
                     found.append(view)
+                    found_ids.add(view_id)
```
doctrino, line 612: Recursion bug.

Juliamg (approved):
Lgtm. I have run through the tutorial as well. A small insignificant note is if the `.env` file that is passed to the client should be renamed here

Screenshot 2024-12-16 at 09 18 56

PR #847, opened 2024-12-16
[NEAT-642] 🧑‍🎨 Neat Rule Store, Part2

doctrino:


tests/tests_unit/test_store/tests_rules_store.py
```diff
@@ -0,0 +1,23 @@
+import yaml
+from pytest_regressions.data_regression import DataRegressionFixture
+
+from cognite.neat._rules import catalog, exporters, importers, transformers
+from cognite.neat._store import NeatRulesStore
+
+
+class TestRuleStore:
+    def test_write_transform_read(self, deterministic_uuid4: None, data_regression: DataRegressionFixture) -> None:
```
Juliamg, line 9: Could you add a test case where there are issues from a read, write or transform step? 

And also check the provenance steps, that `get_last_successful_entity` produces the expected entity?

doctrino: I kept one failing method, as the error logic for `write`, `transform`, `read` is the same.

Juliamg (approved):
Great with the new tests 👍

PR #855, opened 2024-12-19
[NEAT-651] 😃 Increase useability

doctrino:
image


cognite/neat/_rules/models/dms/_rules_input.py
```diff
@@ -289,6 +290,14 @@ def dump(self) -> dict[str, Any]:
             "Nodes": [node_type.dump(default_space) for node_type in self.nodes or []] or None,
         }
 
+    @classmethod
```
Juliamg, line 293: why not make this a property as well?

doctrino: Classmethods cannot be properties

Juliamg (approved):

PR #864, opened 2025-01-02
[NEAT-638] 📖 Support reading from URL

doctrino:


cognite/neat/_utils/reader/_base.py
```diff
@@ -133,6 +141,12 @@ def exists(self) -> bool:
         response = requests.head(self._url)
         return 200 <= response.status_code < 400
 
+    def materialize_path(self) -> Path:
```
nikokaoja, line 144: this works also for XLSX, which is binary?

nikokaoja (approved):
see my comment for binary files from URL

PR #867, opened 2025-01-03
[NEAT-660] 🏄 Refactor catch issues contextmanager

doctrino:
Realization that this  could be simplified.

### DIFF
--- a/cognite/neat/_rules/transformers/_verification.py
+++ b/cognite/neat/_rules/transformers/_verification.py
@@ -1,7 +1,7 @@
 from abc import ABC
 
 from cognite.neat._client import NeatClient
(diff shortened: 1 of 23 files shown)


cognite/neat/_rules/transformers/_verification.py
```diff
@@ -39,8 +39,7 @@ def transform(self, rules: T_ReadInputRules) -> T_VerifiedRules:
         verified_rules: T_VerifiedRules | None = None
         # We need to catch issues as we use the error args to provide extra context for the errors/warnings
         # For example, which row in the spreadsheet the error occurred o
-        issues = IssueList()
```
nikokaoja, line 42: I like this , as previously we were forced to init IssueList, this looks more elegant

nikokaoja (approved):

PR #884, opened 2025-01-07
[NEAT-669] 🫣Continue when direct relation limit is reached. 

doctrino:


nikokaoja (approved):
Suggestion: if possible add to warning that which relationship will be kept and which preserved is not deterministic. 

Suggestion: to have deterministic  cut out of direct relation, sort list of direct relations prior cut.

PR #893, opened 2025-01-09
[NEAT-679] 🛑 Warning for lack of acccess

doctrino:
## Before
image

## After
image

### DIFF
--- a/cognite/neat/_store/_graph_store.py
+++ b/cognite/neat/_store/_graph_store.py
@@ -15,6 +15,7 @@ from cognite.neat._graph._shared import rdflib_to_oxi_type
 from cognite.neat._graph.extractors import RdfFileExtractor, TripleExtractors
 from cognite.neat._graph.queries import Queries
 from cognite.neat._graph.transformers import Transformers
+from cognite.neat._issues import IssueList, catch_issues
 from cognite.neat._rules.analysis import InformationAnalysis
 from cognite.neat._rules.models import InformationRules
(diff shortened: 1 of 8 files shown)


cognite/neat/_store/_graph_store.py
```diff
@@ -157,39 +158,41 @@ def from_oxi_store(cls, storage_dir: Path | None = None, rules: InformationRules
...
-            )
-            for activity in activities:
-                self.provenance.append(
-                    Change.record(
-                        activity=activity,
-                        start=_start,
-                        end=_end,
-                        description=f"Extracted triples to graph store using {type(extractor).__name__}",
-                    )
+    def write(self, extractor: TripleExtractors) -> IssueList:
```
doctrino, line 161: The graph store has no place to store issues, and we do not have an inspect command. Open question, how to solve this?

nikokaoja: Why not adding self.issue_list to NeatGraphStore upon init, which can be updated (append issue) from each extractor?

We providing issue list when loading RDF file: https://github.com/cognitedata/neat/blob/4a1ff6d4cb078dbeb91173cc0dd914d5aa46c455/cognite/neat/_graph/extractors/_rdf_file.py#L29

nikokaoja (changes requested):
Suggest moving catching issues at extractor level as we are doing this when loading RDFFile

nikokaoja (approved):

PR #900, opened 2025-01-09
Upgrade Split Multi Value Property Transfomer

nikokaoja:


cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -1,79 +1,89 @@
...
+                                               <{unknownType}>)) AS ?valueType)
+                   }}
+
+                   GROUP BY ?subjectType ?property
+                   HAVING (COUNT(DISTINCT ?valueType) > 1)"""
+
+        return query.format(unknownType=self.unknown_type)
+
+    def _count_query(self) -> str:
+        query = """SELECT (COUNT(*) AS ?tripleCount)
```
doctrino, line 50: This is an upper bound?

cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -1,79 +1,89 @@
...
 
-    _unknown_property_template: str = """SELECT ?s ?o WHERE {{
+        return query.format(unknownType=self.unknown_type, subject_uri=type_, property_uri=property_)
 
-                                ?s a <{subject_uri}> .
-                                ?s <{property_uri}> ?o .
-                                FILTER NOT EXISTS {{ ?o a ?objectType }}
-                                }}"""
+    def _iterator(self, graph) -> Iterator:
+        for type_, property_, _ in graph.query(self._iterate_query()):
```
doctrino, line 72: Why do you select `?valueTypes` if it is ignored here?

cognite/neat/_graph/transformers/_base.py
```diff
@@ -65,9 +65,16 @@ def _iterate_query(self) -> str:
         The query to use for extracting target triples from the graph and performing the transformation.
         Returns:
             A query string.
+
+        !!! note "Complex Queries"
+            In majority of cases the query should be a simple SELECT query. However, in case of more complex queries
+            especially where multiple consecutive queries are needed, one can overwrite the ._iterator() method
```
doctrino, line 71: . For example, when you have iterator and a subiteration.

doctrino (approved):
Good, I see the use case for overwriting `_iterator`, typically when you have O(n x m) types of lookup. A few clarifying questions for my understanding.

cognite/neat/_graph/transformers/_value_type.py
nikokaoja: Correct!

cognite/neat/_graph/transformers/_base.py
nikokaoja: WIll add that to the note

cognite/neat/_graph/transformers/_value_type.py
nikokaoja: It is temp variable that allows selecting pairs of class-properties which two or more value types. I will see to drop it all toghether

PR #902, opened 2025-01-10
[NEAT-684] 🤗 Remove annoyance

doctrino:
### Before
image

### After
image

**Note**: The user should not care about which version of the PySDK they are using. We always have a lower bound that enables all the functionality we need in neat.

### Before
image

### After
image

**Open Question**: I made the `nordic44` a method instead of property. This is for consistency with other methods, but it is up for discussion. Note that the session wrapper that deals with errors do not handle properties.


rsjr (approved):
LGTM

PR #905, opened 2025-01-13
Discovery oxi remote store

nikokaoja:
- Added folder `./scripts/oxi_remote` where docker compose and lightweight script on how to use it is located
- Renamed `NeatGraphStore.from_oxi_store` to `NeatGraphStore.from_oxi_local_store` to distinguish usage of local and remote oxi store
- Added `NeatGraphStore.from_oxi_remote_store`

Preview of usage of remote store:
```
from cognite.neat._store import NeatGraphStore
from cognite.neat import NeatSession


# Assumes that the server runs on localhost port 7878
store = NeatGraphStore.from_oxi_remote_store(endpoint_url="http://localhost:7878")

neat = NeatSession()

# While we are experimenting Oxi remote store this is the way to set it
neat._state.instances._store = store

# use the neat session as usual, e.g.:
neat.read.rdf.examples.nordic44()

```


scripts/oxi_remote/docker-compose.yaml
```diff
@@ -0,0 +1,11 @@
+version: '3.8'
```
rsjr, line 1: Maybe you can specify the volume in the compose file (kinda good practice)

```
version: '3.8'

services:
  oxigraph:
    image: oxigraph/oxigraph
    init: true
    volumes:
      - oxigraph_data:/data
    ports:
      - "7878:7878"
    restart: always

volumes:
  oxigraph_data:
    name: oxigraph_data

```

cognite/neat/_store/_graph_store.py
```diff
@@ -160,8 +160,24 @@ def from_sparql_store(
         return cls(graph, rules)
 
     @classmethod
-    def from_oxi_store(cls, storage_dir: Path | None = None, rules: InformationRules | None = None) -> "Self":
-        """Creates a NeatGraphStore from an Oxigraph store."""
+    def from_oxi_remote_store(
+        cls,
+        query_endpoint: str | None = None,
```
rsjr, line 165: For query/update, are the endpoint suffixes always the same?
If they are, you could have only one parameter like "remote_url", and split the endpoints later on when you call it

nikokaoja: Good point! For Oxi they are !

scripts/oxi_remote/docker-compose.yaml
nikokaoja: Good point!

PR #907, opened 2025-01-13
[NEAT-688] 😁 Classic knowledge graph extractor

doctrino:
image


cognite/neat/_rules/transformers/_converters.py
```diff
@@ -815,6 +815,69 @@ def description(self) -> str:
...
+
+    def __init__(self, instance_namespace: Namespace) -> None:
+        self.instance_namespace = instance_namespace
+
+    @property
+    def description(self) -> str:
+        return "Update the classic data model to the data types in Cognite Core."
+
+    def transform(self, rules: InformationRules) -> InformationRules:
+        output = rules.model_copy(deep=True)
```
nikokaoja, line 837: Beyond this release and PR we should try to come up with similar form for modifying rules like we are doing with triples, i.e. what row to remove, what to add , etc.

PR #910, opened 2025-01-14
[NEAT-675] 🙂 Remove namespace literals

doctrino:


tests/tests_integration/test_session/test_read.py
```diff
@@ -46,15 +46,18 @@ def test_read_pump_hello_world(self, cognite_client: CogniteClient) -> None:
...
+        try:
+            neat.to.session(session_file)
 
             neat2 = NeatSession()
-            neat2.read.session(session_file.name)
+            neat2.read.session(session_file)
 
             assert (neat2._state.instances.store.graph - neat._state.instances.store.graph).serialize() == "\n"
+        finally:
+            session_file.unlink()
```
doctrino, line 63: This test failed on Windows.  No it works

tests/tests_unit/graph/test_loaders/test_dms_loader.py
```diff
@@ -70,7 +70,7 @@ def test_imf_attribute_nodes():
     knowledge_nodes = list(loader.load())
 
     assert len(knowledge_nodes) == 56
-    assert knowledge_nodes[0].sources[0].properties["predicate"].startswith("http")
+    assert knowledge_nodes[0].sources[0].properties["predicate"] == "CFIHOS-40000524"
```
doctrino, line 73: This is a result of the new change. All the IMF properties are AnyURI in the information model, which gets translated into text fields, and thus removes the namespace. 
image

Open question: Should AnyURI instead translate into a connection with unknown source? In DMS, a direct relation that has not specified source.

nikokaoja (approved):
Approving ... though we solved this closer to the source in the latest PRs

PR #911, opened 2025-01-14
☢ Multi source neat store

nikokaoja:
## Added

- Support RDF Datasets in NeatGraphStore enabling writing of sources triples to dedicated named graphs

Example:

Screenshot 2025-01-14 at 17 54 22


doctrino (approved):
If I understand this one correctly, you essentially replace the earlier graph object with dataset. Then, everywhere you query the graph you now have an optional URIref you can use to specify which dataset graph to query?

PR #915, opened 2025-01-15
Fix issues with removal of namespace

nikokaoja:


cognite/neat/_graph/queries/_base.py
```diff
@@ -229,16 +229,21 @@ def describe(
                 "null",
             ]:
                 continue
-
+            print(predicate, object_)
```
doctrino, line 232:
```suggestion
```

cognite/neat/_graph/queries/_base.py
```diff
@@ -279,6 +292,7 @@ def describe(
                     # we should not have multiple rdf:type values
                     continue
         if property_values:
+            print(identifier, property_values)
```
doctrino, line 295:
```suggestion
```

session.zip
doctrino, line 1: What is this? this looks like something that should not be in the commit.

doctrino (commented):
some minor cleanup

doctrino (approved):

PR #921, opened 2025-01-16
😵‍💫 First pass on handling out of neat session modifications on rules

nikokaoja:
Importing rules which were exported then modified externally then re-imported to the neat session now are passing through  `_sync_import` method in rules_store

Not all use-case are implemented, it is a first pass


tests/tests_integration/test_session/test_data_model_to.py
```diff
@@ -81,3 +86,39 @@ def test_to_cdf_recreate_drop_data(self, neat_client: NeatClient) -> None:
...
+        with pytest.raises(NeatValueError) as e:
+            neat._state.rule_import(importers.ExcelImporter(tmp_path / "pump_example.xlsx"))
+
+        assert (
+            "Imported rules and rules which were used as the source for"
+            " them and which are are already in "
+            "this neat session have the same data model id"
+        ) in e.value.raw_message
+
+    def test_stopping_loading_model_source_mode_not_in(self, neat_client: NeatClient, tmp_path: Path) -> None:
```
nikokaoja, line 112:
```suggestion
    def test_stopping_loading_model_which_source_is_not_in_session(self, neat_client: NeatClient, tmp_path: Path) -> None:
```

doctrino (approved):

PR #922, opened 2025-01-17
[NEAT-704] 🤨 Support external ID

doctrino:


cognite/neat/_graph/extractors/_classic_cdf/_base.py
```diff
@@ -72,6 +75,8 @@ class ClassicCDFBaseExtractor(BaseExtractor, ABC, Generic[T_CogniteResource]):
         camel_case (bool, optional): Whether to use camelCase instead of snake_case for property names.
             Defaults to True.
         as_write (bool, optional): Whether to use the write/request format of the items. Defaults to False.
+        prefix (str, optional): A prefix to add to the rdf type. Defaults to None.
+        identifier (Literal["id", "externalId"], optional): The identifier to use. Defaults to "id".
```
nikokaoja, line 79: should we call this:
internalId and externalId ? 
No strong opinion

cognite/neat/_graph/extractors/_classic_cdf/_base.py
```diff
@@ -90,6 +95,7 @@ def __init__(
         camel_case: bool = True,
         as_write: bool = False,
         prefix: str | None = None,
+        identifier: typing.Literal["id", "externalId"] = "id",
```
nikokaoja, line 98: suggestion as this is config maybe we call this: identifier_type , so it is clear it is not actually the identifier?

nikokaoja: or maybe short id_type

cognite/neat/_graph/extractors/_classic_cdf/_base.py
```diff
@@ -154,10 +183,22 @@ def _item2triples_special_cases(self, id_: URIRef, dumped: dict[str, Any]) -> li
         """This can be overridden to handle special cases for the item."""
         return []
 
+    @staticmethod
+    def _external_id_as_uri_suffix(external_id: str | None) -> str:
+        if external_id == "":
```
nikokaoja, line 188: My worry here is if there are multiple assets which have no external id, we would slam a lot of triples to same subject.

We should capture on extraction issues , maybe some sort of counter whenever we end up in any of this listed edge cases.

cognite/neat/_graph/extractors/_classic_cdf/_base.py
```diff
@@ -182,9 +223,27 @@ def _get_rdf_type(self, item: T_CogniteResource) -> str:
 
     def _as_object(self, raw: Any, key: str) -> Literal | URIRef:
         if key in {"data_set_id", "dataSetId"}:
-            return self.namespace[f"{InstanceIdPrefix.data_set}{raw}"]
+            if self.identifier == "externalId" and self.lookup_dataset_external_id:
+                try:
+                    data_set_external_id = self.lookup_dataset_external_id(raw)
+                except KeyError:
+                    return Literal("Unknown data set", datatype=XSD.string)
```
nikokaoja, line 230:
```suggestion
                    return Literal("Unknown data set")
```
it will automatically pick up xsd type

cognite/neat/_graph/extractors/_classic_cdf/_base.py
```diff
@@ -182,9 +223,27 @@ def _get_rdf_type(self, item: T_CogniteResource) -> str:
...
+                else:
+                    return self.namespace[
+                        f"{InstanceIdPrefix.data_set}{self._external_id_as_uri_suffix(data_set_external_id)}"
```
nikokaoja, line 233: I think we use `_` between prefix for type and actual id so:

```
f"{InstanceIdPrefix.data_set}_{self._external_id_as_uri_suffix(data_set_external_id)}"
```

cognite/neat/_graph/extractors/_classic_cdf/_base.py
```diff
@@ -182,9 +223,27 @@ def _get_rdf_type(self, item: T_CogniteResource) -> str:
 
     def _as_object(self, raw: Any, key: str) -> Literal | URIRef:
```
nikokaoja, line 224:
```suggestion
    def _as_object(self, raw: Any, key: str) -> Literal | URIRef:
        """Return properly formatted object part of s-p-o triple"""
```

Old guy needs to understand it is object of triple not python object :)

nikokaoja (approved):
My main comments:

- name of config arg for identifier
- capturing external ids edge cases (e.g., when it is missing, etc.)

cognite/neat/_graph/extractors/_classic_cdf/_base.py
doctrino: That is already included in the prefix lookup. `InstanceIdPrefix.data_set` = `DataSet_`

cognite/neat/_graph/extractors/_classic_cdf/_base.py
doctrino: I agree. Adding some warning here, that should be captured higher up in the stack.

PR #923, opened 2025-01-18
[NEAT-706] 😇 Keep reference

doctrino:


cognite/neat/_session/_read.py
```diff
@@ -181,6 +181,18 @@ def graph(
...
+        )
+
+    def _graph(
+        self,
+        root_asset_external_id: str,
+        limit_per_type: int | None = None,
+        identifier: Literal["id", "externalId"] = "id",
+        reference_timeseries: bool = False,
+        reference_files: bool = False,
+    ) -> IssueList:
```
doctrino, line 195: Small trick to avoid exposing these two new parameters.

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -823,11 +823,21 @@ class ClassicPrepareCore(RulesTransformer[InformationRules, InformationRules]):
     - ClassicTimeseries.isString from boolean to string
     - Add class ClassicSourceSystem, and update all source properties from string to ClassicSourceSystem.
     - Rename externalId properties to classicExternalId
-    - Renames the Relationship.sourceExternaId and Relationship.targetExternalId to startNode and endNode
+    - Renames the Relationship.sourceExternalId and Relationship.targetExternalId to startNode and endNode
```
nikokaoja, line 826: Great !

nikokaoja: Beware that we have been "extending" XSD typing for timeseries before as you can see here:

https://github.com/cognitedata/neat/blob/817bda65f437a575024b1e67ae140e85d5456065/cognite/neat/_graph/extractors/_iodd.py#L246

which we are able to handle both in inference as well when converting from Info to DMS rules.

nikokaoja (approved):

PR #927, opened 2025-01-20
[NEAT-714] 🏎Speedup graph extraction

doctrino:


tests/tests_integration/test_session/test_graph_flow/test_classic_to_dms.yml
```diff
@@ -1833,9 +1833,9 @@ instances:
     externalId: ClassicLabel
     space: sp_windfarm
 - endNode:
-    externalId: MetMast
+    externalId: http://purl.org/cognite/cdf-classic#Asset_8279627777657373
```
doctrino, line 1836: Something goes wrong here. This should be `Asset_8279627777657373`

PR #928, opened 2025-01-21
🤕 Enable NEAT type 

nikokaoja:
- auto-select the most optimum graph store
- enable neat type in addition to rdf type, where neat type is prioritized
- set form for building neat internal data model


nikokaoja (commented):
Introducing neat:type brought in complexity in all our queries.

My suggestion, ditching this PR, and re-writing transformer to update rdf:type instead of adding neat:type

PR #935, opened 2025-01-23
💁Create API

doctrino:
Moving from the .prepare to new .create API

### DIFF
--- a/cognite/neat/_session/_create.py
+++ b/cognite/neat/_session/_create.py
@@ -1,0 +1,144 @@
+from typing import Literal
+
+from cognite.client.data_classes.data_modeling import DataModelIdentifier
+
+from cognite.neat._issues import IssueList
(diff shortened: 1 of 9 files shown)


cognite/neat/_session/_create.py
```diff
@@ -0,0 +1,144 @@
...
+    ToEnterpriseModel,
+    ToSolutionModel,
+)
+
+from ._state import SessionState
+from .exceptions import NeatSessionError, session_class_wrapper
+
+
+@session_class_wrapper
+class CreateAPI:
```
nikokaoja, line 20: Assuming we will not have `neat.create.instance`, this is fine with me.

nikokaoja (approved):

cognite/neat/_session/_create.py
doctrino: Yes, I did the same thinking.

PR #936, opened 2025-01-24
😅Solution composition

doctrino:


cognite/neat/_rules/models/_base_rules.py
```diff
@@ -326,19 +328,20 @@ def dump(
...
-            value = getattr(self, field_name)
-            # Ensure deterministic order of properties, classes, views, and so on
-            if isinstance(value, SheetList):
-                value.sort(key=lambda x: x._identifier())
+        if sort:
+            for field_name in self.model_fields.keys():
+                value = getattr(self, field_name)
+                # Ensure deterministic order of properties, classes, views, and so on
+                if isinstance(value, SheetList):
+                    value.sort(key=lambda x: x._identifier())
```
doctrino, line 336: Introduced this as a parameter and default to not sorting. The motivation is that when dumping a solution model is that you want to have a special order of properties such that the 'Enterprise' view properties is next to the one that points to it.

cognite/neat/_session/_create.py
```diff
@@ -60,24 +58,23 @@ def enterprise_model(
                 new_model_id=data_model_id,
                 org_name=org_name,
                 dummy_property=dummy_property,
-                move_connections=move_connections,
+                move_connections=True,
```
nikokaoja, line 61: great !

nikokaoja (approved):

PR #938, opened 2025-01-24
😎Inference using subClassOf

doctrino:


cognite/neat/_constants.py
```diff
@@ -86,7 +86,6 @@ class NEAT(DefinedNamespace):
     _fail = True
     _NS = Namespace("http://thisisneat.io/internal/")
 
-    type: URIRef  # type property used to express a type of a subject
```
doctrino, line 89: No more `Neat.type` :)

cognite/neat/_constants.py
```diff
@@ -86,8 +86,8 @@ class NEAT(DefinedNamespace):
...
     UnknownType: URIRef  # Unknown type used to express that the type of a subject is unknown
+    EmptyType: URIRef  # Empty type used to express that the type of a subject is empty
```
nikokaoja, line 90: Good catch for this one, as we might hit edge case where a type of subject is not set... which indicates bad graph

cognite/neat/_graph/transformers/_value_type.py
```diff
@@ -352,7 +356,10 @@ def operation(self, query_result_row: ResultRow) -> RowTransformationOutput:
         if self.drop_property:
             row_output.remove_triples.append((instance, self.subject_predicate, object_literal))
 
-        row_output.add_triples.append((instance, NEAT.type, self._namespace[str(object_literal.toPython())]))
+        row_output.remove_triples.append((instance, RDF.type, self.subject_type))
+        new_type = self._namespace[quote(object_literal.toPython())]
+        row_output.add_triples.append((instance, RDF.type, new_type))
+        row_output.add_triples.append((new_type, RDFS.subClassOf, self.subject_type))
```
nikokaoja, line 362: Strong suggestion to update type of `add_triples` and `remove_triples` from list to set, as there is no point in holding duplicated triples.

nikokaoja (approved):
Nice!

cognite/neat/_graph/transformers/_value_type.py
doctrino: Very nice catch. Changing it

PR #939, opened 2025-01-24
🥏 Public interface for on disc oxi

nikokaoja:
Main issue, I cannot get a nice printout of error when oxi is locked
Screenshot 2025-01-24 at 16 29 29

it is suppose to print out error and fix that are defined in `OxigraphStorageLockedError`

### DIFF
--- a/cognite/neat/_session/_base.py
+++ b/cognite/neat/_session/_base.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from typing import Literal
 
 from cognite.client import CogniteClient
@@ -81,12 +82,14 @@ class NeatSession:
         self,
         client: CogniteClient | None = None,
         storage: Literal["memory", "oxigraph"] | None = None,
+        storage_path: str | None = None,
         verbose: bool = True,
         load_engine: Literal["newest", "cache", "skip"] = "cache",
     ) -> None:
         self._verbose = verbose
         self._state = SessionState(
             store_type=storage or self._select_most_performant_store(),
+            storage_path=Path(storage_path) if storage_path else None,
             client=NeatClient(client) if client else None,
         )
         self.read = ReadAPI(self._state, verbose)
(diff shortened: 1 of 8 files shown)


cognite/neat/_store/_graph_store.py
```diff
@@ -685,3 +681,8 @@ def _shorten_summary(self, summary: pd.DataFrame) -> pd.DataFrame:
     @property
     def named_graphs(self) -> list[URIRef]:
         return [cast(URIRef, context.identifier) for context in self.dataset.contexts()]
+
+    @property
+    def empty(self) -> bool:
+        """Cheap way to check if the graph store is empty."""
+        return self.queries.has_data()
```
doctrino, line 688: Shouldn't this be
```suggestion
    def empty(self) -> bool:
        """Cheap way to check if the graph store is empty."""
        return not self.queries.has_data()
```
That is the store is empyt if it does not have data?

cognite/neat/_session/_base.py
```diff
@@ -270,11 +273,15 @@ def _repr_html_(self) -> str:
...
 
-        if state.instances.has_store:
+        if state.instances.empty:
             output.append(f"<H2>Instances</H2> {state.instances.store._repr_html_()}")
 
         return "<br />".join(output)
 
+    def close(self) -> None:
+        """Close the session and release resources."""
+        self._state.instances.store.dataset.close()
```
doctrino, line 283: We are putting more responsibility on the user with this. I think in addition to this method, we should call it `__del__` method.

Then when you run the following

```python
neat = NeatSession()
... # doing some stuff

# This creates a new NeatSession and thus we lose all reference so the above `NeatSession`
# When the memory allocation for the previous object is cleaned up by the garbage collector we 
# want to automatically close to session. Note this does not work if the user suddenly shuts down the kernel, 
# but better than nothing.
neat = NeatSession()
```

```suggestion
    def close(self) -> None:
        """Close the session and release resources."""
        self._state.instances.store.dataset.close()

    def __del__(self) -> None:
        """Close the session and release resources."""
        self.close()
```

doctrino (approved):
In addition to the comments, we should we also make `NeatSession` a context object?

```python
with NeatSession() as neat:
    ... # Do stuff

```
Which automatically calls `.close` upon exit. This will not be used by the average user, but advanced users will benefit from it.

PR #942, opened 2025-01-26
[NEAT-724]📖 Update tutorials

doctrino:
Realized that `.convert('dms')` does not make sense anymore, so now `.convert()` will always convert `Info -> DMS`. It is not possible to do `DMS -> Info`. 

In addition exposed the parameter of what to do with reserved properties `convert(reserved_properties="error/skip")`.


cognite/neat/_session/_show.py
```diff
@@ -98,7 +98,7 @@ def __init__(self, state: SessionState) -> None:
         self.implements = ShowDataModelImplementsAPI(self._state)
 
     def __call__(self) -> Any:
-        if not self._state.rule_store.empty:
+        if self._state.rule_store.empty:
```
doctrino, line 101: Bug

cognite/neat/_session/_base.py
```diff
@@ -150,35 +148,24 @@ def verify(self) -> IssueList:
         print("This action has no effect. Neat no longer supports unverified data models.")
         return IssueList()
 
-    def convert(self, target: Literal["dms", "information"]) -> IssueList:
+    def convert(self, reserved_properties: Literal["error", "skip"] = "skip") -> IssueList:
```
doctrino, line 151: Should we add a warning if the user set `reserved_propertie="dms")`, i.e., do not update how they call this method.

nikokaoja: reserved_properties = ["warning", "error"]

nikokaoja (approved):

PR #945, opened 2025-01-27
Expose set replace type

doctrino:


nikokaoja (approved):
maybe rename it to

neat.set.instances.type_using_property("NeatIncAsset", "assetCategory")

As it nicely reads as:

"Neat sets instances type using property (...)"

WDYT?

PR #946, opened 2025-01-27
Update prepare prefix

nikokaoja:
NeatSession.prepare.data_model.prefix works only on verified Information and DMS rules, where in case of both rules are provided, it will use DMS rules


cognite/neat/_rules/transformers/_converters.py
```diff
@@ -157,73 +149,94 @@ def _fix_properties(cls, definitions: SheetList[InformationProperty]) -> SheetLi
...
+        return f"Prefixes all entities with {self._prefix!r} prefix if they are in the same space as data model."
 
-    def transform(self, rules: ReadRules[T_InputRules]) -> ReadRules[T_InputRules]:
-        in_ = rules.rules
-        if in_ is None:
-            return rules
-        copy: T_InputRules = dataclasses.replace(in_)
-        if isinstance(copy, InformationInputRules):
-            prefixed_by_class: dict[str, str] = {}
+    def transform(self, rules: InformationRules | DMSRules) -> InformationRules | DMSRules:
```
doctrino, line 162: You will either do information->Infomrtion, or DMS->DMS
```suggestion
    def transform(self, rules: T_VerifiedIn) -> T_VerifiedIn:
```

Maybe an even better alternative use overload

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -157,73 +149,94 @@ def _fix_properties(cls, definitions: SheetList[InformationProperty]) -> SheetLi
         return fixed_definitions
 
 
-class PrefixEntities(RulesTransformer[ReadRules[T_InputRules], ReadRules[T_InputRules]]):  # type: ignore[type-var]
-    """Prefixes all entities with a given prefix."""
+class PrefixEntities(ConversionTransformer):  # type: ignore[type-var]
```
doctrino, line 152:
```suggestion
class PrefixEntities(ConversionTransformer[T_VerifiedIn, T_VerifiedOut]): 
```

doctrino (approved):
Suggest overloading the .transform(). Otherwise all good.

cognite/neat/_rules/transformers/_converters.py
nikokaoja: Didn't I set up overload correctly ? hm ... that is strange

PR #948, opened 2025-01-28
🚧 Enable manual transformations of rules outside of NeatSession

nikokaoja:
image


cognite/neat/_store/_rules_store.py
```diff
@@ -66,63 +73,153 @@ def calculate_provenance_hash(self, shorten: bool = True) -> str:
             return calculated_hash[:8]
         return calculated_hash
 
-    def import_rules(
-        self, importer: BaseImporter, validate: bool = True, client: NeatClient | None = None
-    ) -> IssueList:
-        if self.empty:
-            return self._import_rules(importer, validate, client)
+    def _rules_import_verify_convert(
```
nikokaoja, line 76: I've had to re-shuffle things a bit to enable manual import/transformation

cognite/neat/_store/_rules_store.py
```diff
@@ -66,63 +73,153 @@ def calculate_provenance_hash(self, shorten: bool = True) -> str:
...
-        if self.provenance:
-            raise NeatValueError(f"Data model already exists. Cannot import {agent_tool.source_uri}.")
-        return self.do_activity(action, agent_tool)
+        if self.empty:
+            return self.do_activity(
+                partial(self._rules_import_verify_convert, importer, validate, client),
+                importer,
+            )
+        else:
+            print("Manual transformation of rules is alpha feature. Use with caution.")
```
doctrino, line 210: I think we should avoid print in the rule store if we can. Instead, it should be done one the session level

doctrino: I wonder if we should pass en extra argument in read.excel() to enable this feature. Making the user explicit aware that they are using an alpha feature?

doctrino (approved):
I think you did some good splitting in the rules store, but it made it slightly harder to see the changes of this PR. I think I got it. I like the idea of putting this behind an alpha flag.

PR #964, opened 2025-02-01
[NEAT-739] 🙂 Handle bad names

doctrino:


cognite/neat/_rules/models/_rdfpath.py
```diff
@@ -327,6 +328,7 @@ def parse_table_lookup(raw: str) -> TableLookup:
 
 
 def parse_rule(rule_raw: str, rule_type: TransformationRuleType | None) -> RDFPath:
+    rule_raw = urllib.parse.unquote(rule_raw)
```
doctrino, line 331: Up for discussion how to deal with this. The old RDFPath is more restrictive than a URI, which I think might be a problem. 

Should we reduce the RDFPath to the bare minimum, no hops allowed before v1 and instead reintrodure the more advanced concepts with an alpha flag?

nikokaoja: Fully agree, everything that is doing hops, etc. should be neat.prepare.instances....

nikokaoja (approved):

PR #975, opened 2025-02-05
[NEAT-741] 💼 Wrapping Issues

doctrino:
## Error property
image

## Error metadata
image


cognite/neat/_rules/models/dms/_validation.py
```diff
@@ -54,13 +55,19 @@ class DMSValidation:
...
+        client: NeatClient | None = None,
+        read_info_by_spreadsheet: dict[str, SpreadsheetRead] | None = None,
+    ) -> None:
         self._rules = rules
         self._client = client
         self._metadata = rules.metadata
         self._properties = rules.properties
         self._containers = rules.containers
         self._views = rules.views
+        self._read_info_by_spreadsheet = read_info_by_spreadsheet or {}
```
doctrino, line 70: Adjust row numbers directly in the Validtion, instead of using the context manager. Making it more explicit.

tests/tests_unit/rules/test_importers/test_excel_importer.py
```diff
@@ -34,15 +35,11 @@ def invalid_rules_filepaths():
...
+                PropertyValueError(
                     row=5,
-                    type="bool_parsing",
-                    msg="Input should be a valid boolean, unable to interpret input",
-                    input="Apple",
-                    url=f"https://errors.pydantic.dev/{major}.{minor}/v/bool_parsing",
-                )
+                    column="Is List",
+                    error=NeatValueError("Expected a bool type, got 'Apple'"),
+                ),
```
doctrino, line 54: Example of improved error.

cognite/neat/_issues/_base.py
```diff
@@ -530,38 +331,3 @@ def _get_subclasses(cls_: type[T_Cls], include_base: bool = False) -> Iterable[t
...
-        error_args: Additional arguments to pass to the error class. The only use case as of (2025-01-03) is to pass
-            the read_info_by_sheet to the error class such that the row numbers can be adjusted to match the source
-            spreadsheet.
-
-    Returns:
-        IssueList: The list of issues.
-
-    """
-    with catch_warnings() as issues:
-        try:
```
doctrino, line 562: Moved out such that they can depend on the factory functions. If not we would get circular dependencies.

cognite/neat/_issues/_base.py
```diff
@@ -215,207 +208,34 @@ def __eq__(self, other: object) -> bool:
...
 
+    """
 
-T_NeatIssue = TypeVar("T_NeatIssue", bound=NeatIssue)
+    def __init__(self, errors: Sequence[NeatIssue]):
+        self.errors = IssueList(errors)
 
 
-class NeatIssueList(list, Sequence[T_NeatIssue], ABC):
+class IssueList(list, Sequence[NeatIssue]):
```
doctrino, line 233: No more generic `NeatIssueList`. This was never used, instead we only have the `IssueList`. This means we can no longer create a list of specific types of issues, but that was never used.

cognite/neat/_issues/_base.py
```diff
@@ -215,207 +208,34 @@ def __eq__(self, other: object) -> bool:
...
 
-    @classmethod
-    def from_errors(cls, errors: "list[ErrorDetails | NeatError]", **kwargs) -> "list[NeatError]":
-        """Convert a list of pydantic errors to a list of Error instances.
-
-        This is intended to be overridden in subclasses to handle specific error types.
-        """
-        all_errors: list[NeatError] = []
-        read_info_by_sheet = kwargs.get("read_info_by_sheet")
-
```
doctrino, line 226: All factory methods moved to its own module to avoid circular dependencies and adhere to single responsibility. This module now only have base classes, while factory module is for creating issues.

nikokaoja (approved):

PR #977, opened 2025-02-06
[NEAT-756] 🥰 Combine info and analysis

doctrino:
Moved all analysis method into a shared class. Removed the use of generics, and instead do most analysis of the Information rules. Skipped all methods that are not used in the code base.

I think as we continue to work on the Graph Store and DMSLoader, we will replaces some of the methods, like the ones using `NeatId`. Ideally, neat id should not be exposed to the store.


cognite/neat/_rules/analysis/_base.py
```diff
@@ -65,271 +61,258 @@ def to_pandas(self) -> pd.DataFrame:
...
+        return parents_by_class
 
-        return class_property_pairs
+    @staticmethod
+    def _include_ancestors(parents_by_class: dict[T_Hashable, set[T_Hashable]]) -> None:
+        # Topological sort to ensure that classes include all ancestors
+        for class_entity in list(TopologicalSorter(parents_by_class).static_order()):
+            parents_by_class[class_entity] |= {
+                grand_parent for parent in parents_by_class[class_entity] for grand_parent in parents_by_class[parent]
+            }
```
doctrino, line 117: Not the last method did not do this. That could lead to only including the immediate parent. This guarantees that all parents are included.

nikokaoja (approved):

PR #982, opened 2025-02-07
[NEAT-745] 🫵 feat; Support setting reference model directly

doctrino:
image

### DIFF
--- a/cognite/neat/_session/_to.py
+++ b/cognite/neat/_session/_to.py
@@ -5,14 +5,18 @@ from pathlib import Path
 from typing import Any, Literal, overload
 
 from cognite.client import data_modeling as dm
+from cognite.client.data_classes.data_modeling import DataModelIdentifier
 
 from cognite.neat._alpha import AlphaFlags
 from cognite.neat._constants import COGNITE_MODELS
 from cognite.neat._graph import loaders
+from cognite.neat._issues import IssueList, catch_issues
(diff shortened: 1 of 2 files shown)


cognite/neat/_session/_to.py
```diff
@@ -71,19 +76,45 @@ def excel(
...
+                    return issues
+            if ref_rules is not None:
                 prefix = "Ref"
-            reference_rules_with_prefix = self._state.last_reference, prefix
+                if (
+                    isinstance(ref_rules.metadata, DMSMetadata)
+                    and ref_rules.metadata.as_data_model_id() in COGNITE_MODELS
+                ):
+                    prefix = "CDM"
+                reference_rules_with_prefix = ref_rules, prefix
```
doctrino, line 117: This is something very special for writing to Excel, so kept the logic here.

cognite/neat/_session/_to.py
```diff
@@ -35,17 +39,18 @@ def __init__(self, state: SessionState, verbose: bool) -> None:
     def excel(
         self,
         io: Any,
-        include_reference: bool = True,
+        include_reference: bool | DataModelIdentifier = True,
```
doctrino, line 42: I think this is small feature which does not have to be behind an alpha flag.

nikokaoja (approved):

PR #984, opened 2025-02-10
☔ Neat 757 workshop learning del cells row

nikokaoja:
- skip rows in Excel rules if all values are None except for internal columns


doctrino (commented):
I think this is risky, need some sort of user feedback to tell that row was dropped.

doctrino (approved):
I was expecting a warning, but if I read it correctly you only drop the rows that does not contain anything critical? so it is ok to silently drop?

PR #990, opened 2025-02-17
[NEAT-753] 🥸 Handle duplicated property names

doctrino:


cognite/neat/_graph/queries/_base.py
```diff
@@ -168,7 +167,6 @@ def describe(
         instance_id: URIRef,
         instance_type: URIRef | None = None,
         property_renaming_config: dict | None = None,
-        property_types: dict[str, EntityTypes] | None = None,
```
doctrino, line 171: This was not used

nikokaoja (approved):

PR #995, opened 2025-02-17
Neat 780 information prevent duplicated resources

nikokaoja:


cognite/neat/_rules/models/dms/_validation.py
```diff
@@ -150,6 +154,64 @@ def validate(self) -> IssueList:
...
+                        property_.view_property,
+                        "property",
+                        (
+                            "the Properties sheet at row "
+                            f"{properties_sheet.adjusted_row_number(row_no) if properties_sheet else row_no + 1}"
+                            " if data model is read from a spreadsheet."
+                        ),
+                    )
+                )
+            visited.add(property_._identifier())
```
doctrino, line 178: Suggest doing something like this. Then you handle more than 2 duplicates as one error. 
```suggestion
visited = defaultdict(list)
for row_no, property_ in enumerate(self._properties):
    visited[property_._identifier()].append(row_no)

for identifier, row_numbers in visited.items():
    if len(row_numbers) == 1:
        continue
    issue_list.append(
        ResourceDuplicatedError(
            identifier,
            "property",
            (
                "the Properties sheet at row "
                f"{humanize_collection([properties_sheet.adjusted_row_number(row_no) if proerties_sheet for row_no in row_numbers])}"
                " if data model is read from a spreadsheet."
            ),
        )
    )
```

doctrino (approved):
One suggestion

PR #999, opened 2025-02-20
[NEAT-783] 🤕 Fix space

doctrino:

PR #1001, opened 2025-02-20
Neat 790 investigate neat space naming issue

nikokaoja:
- Now DMS rules validation are checked for existence of all components of schema, raise error if at least one is missing
- Enable data model to be composed of views from multiple spaces with warning being raised


cognite/neat/_rules/models/dms/_exporter.py
```diff
@@ -114,7 +115,11 @@ def _create_spaces(
         views: ViewApplyDict,
         data_model: dm.DataModelApply,
     ) -> SpaceApplyDict:
-        used_spaces = {container.space for container in containers.values()} | {view.space for view in views.values()}
+        used_spaces = (
```
nikokaoja, line 118: this was not correct, as only spaces of containers and views were considered, and not also data model space

cognite/neat/_rules/models/dms/_exporter.py
```diff
@@ -114,7 +115,11 @@ def _create_spaces(
...
+            {container.space for container in containers.values()}
+            | {view.space for view in views.values()}
+            | {data_model.space}
+        )
         if len(used_spaces) == 1:
             # We skip the default space and only use this space for the data model
```
doctrino, line 124: This comment is outdated with the change above.

doctrino (approved):
Suggest we take in #999 first as it changes the data in CDF and will lead to tests passing.

doctrino (dismissed):

doctrino (approved):

PR #1009, opened 2025-02-25
[NEAT-808]🐫 Migration tooling to utils

doctrino:
Used in the migration work to do quick comparisons between node and source TimeSeries/Asset/Event/FileMetadata/Sequences.

This is not exposed to the end user and will likely not be yet, but we we need it as we do large scale tests across multiple customers.

image

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


nikokaoja (approved):

nikokaoja (dismissed):

cognite/neat/_utils/migration.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def display_diffs(
+    source_items: typing.Sequence[Asset | TimeSeries | Event | Sequence | FileMetadata],
+    target_view: ViewId | tuple[str, str, str],
+    source: CogniteClient,
+    target: CogniteClient | None = None,
+) -> None:
+    """Displays the differences between the source and target objects."""
+    local_import("deepdiff", "compare")
```
rbtcollins, line 30: Its a poor pattern to call into the import machinery deep within functions in Python - just because we can do it, doesn't mean we should.

There is also a developer affordance impact of having custom import layers - IDEs and other code level tools don't typically handle things that do imports but aren't import statements well.

I do see that you need this in some places, to switch to fallback codepaths, but that isn't the case here for anything in this module. -> If this function called from the contexts where you don't have `deepdiff` available, it will fail at runtime.

So, please either:
- document the failure modes in the docstring, including what it will raise
- or lift the import out of the function so that code which has imported the function knows it is usable.

cognite/neat/_utils/migration.py
```diff
@@ -0,0 +1,154 @@
...
+RESERVED_PROPERTIES = {word.replace("_", "").lower() for word in get_reserved_words("property")}
+
+
+def display_diffs(
+    source_items: typing.Sequence[Asset | TimeSeries | Event | Sequence | FileMetadata],
+    target_view: ViewId | tuple[str, str, str],
+    source: CogniteClient,
+    target: CogniteClient | None = None,
+) -> None:
+    """Displays the differences between the source and target objects."""
```
rbtcollins, line 29:
```suggestion
    """Prints the differences between the source and target objects."""
```

cognite/neat/_utils/migration.py
```diff
@@ -0,0 +1,154 @@
...
+            continue
+        node = node_by_id[item.external_id]
+        classic_node = as_classic(node, item)
+        if classic_node != item:
+            print(f"Failed: {type_} {item.external_id!r}")
+            pprint(DeepDiff(item.dump(), classic_node.dump()))
+
+
+def as_classic(node: Node, classic: T_Classic) -> T_Classic:
+    """Converts a Node to is corresponding Asset/TimeSeries/Event/Sequence/FileMetadata object.
```
rbtcollins, line 62:
```suggestion
    """Converts a Node to its corresponding Asset/TimeSeries/Event/Sequence/FileMetadata object.
```

rbtcollins (changes requested):
This looks like it will be quite nice. However it is effectively not tested today, and if we're depending on it to tell us about the accuracy of the system, it must be.

I've left some thoughts on the structure and patterns as well.

cognite/neat/_utils/migration.py
doctrino: This is an internal function and we do not want to add an extra dependency on DeepDiff to everyone that uses neat, thus we use a local import and mark this as an optional dependency. The local_import function raises an appropriate error if the dependency is missing.

doctrino:
> This looks like it will be quite nice. However it is effectively not tested today, and if we're depending on it to tell us about the accuracy of the system, it must be.
> 
> I've left some thoughts on the structure and patterns as well.

Currently, this is used internally and as you are saying it is not well tested. The thinking is to extend this one as we continue to use it on our pilot customers.

nikokaoja (approved):

nikokaoja (approved):

rbtcollins (approved):
:unicorn:

PR #1011, opened 2025-02-25
⛰Large scale round3

doctrino:
In the DMS Loader: Removal of doing double ingestion due to hierarchical properties, this turns out not to be necessary.


cognite/neat/_graph/loaders/_rdf2dms.py
```diff
@@ -189,33 +187,12 @@ def _load(
...
             for identifier, properties in instance_iterable:
-                yield from self._create_instances(
-                    identifier, properties, projection, stop_on_exception, exclude=it.hierarchical_properties
-                )
-            if it.hierarchical_properties:
-                # Force the creation of instances, before we create the hierarchical properties.
-                yield _END_OF_CLASS
-                yield _START_OF_CLASS(f"{view.external_id} ({humanize_collection(it.hierarchical_properties)})")
-                yield from self._create_hierarchical_properties(it, projection, stop_on_exception)
-            if reader is not instance_iterable:
```
doctrino, line 200: We learned that hierarchical properties are not necessary to account for in the loading into DMS. Thus removing this from the loader.

cognite/neat/_graph/loaders/_rdf2dms.py
```diff
@@ -500,8 +473,6 @@ def _create_instances(
         properties: dict[str | InstanceType, list[Any]],
         projection: _Projection,
         stop_on_exception: Literal[True, False] = False,
-        exclude: set[str] | None = None,
-        include: set[str] | None = None,
```
doctrino, line 504: These were only used for hierarchical properties, thus removing them to simplify.

cognite/neat/_graph/loaders/_rdf2dms.py
```diff
@@ -189,33 +187,12 @@ def _load(
...
-                    identifier, properties, projection, stop_on_exception, exclude=it.hierarchical_properties
-                )
-            if it.hierarchical_properties:
-                # Force the creation of instances, before we create the hierarchical properties.
-                yield _END_OF_CLASS
-                yield _START_OF_CLASS(f"{view.external_id} ({humanize_collection(it.hierarchical_properties)})")
-                yield from self._create_hierarchical_properties(it, projection, stop_on_exception)
-            if reader is not instance_iterable:
+                yield from self._create_instances(identifier, properties, projection, stop_on_exception)
+            if reader is instance_iterable:
```
doctrino, line 191: Small logical bug. We should print if we do not use a progress bar.

nikokaoja (dismissed):
This is great! Avoiding double ingestion speed things up !

doctrino:
Reimplemented as #1022

PR #1016, opened 2025-02-26
[NEAT-796] 🤨 Lost parents

doctrino:
## Before
We dependent on the order of assets extracted. Now we create a triple with all parents we extract. In addition, we are now looking up datasets of parents such that when we use it for space. 

image

## Bump

- [ ] Patch
- [x] Minor
- [ ] Skip

### Added
- Drop-down menus for columns in the Excel representation of DMS Rules (data model)
### Fixed
- The `neat.read.cdf.classic.graph()` no longer loses asset partens when `identifier='externalId'`.


nikokaoja (dismissed):

larshagencognite:
I will review this when it is rebased to master, after merging display name

nikokaoja (approved):

larshagencognite (approved):

PR #1019, opened 2025-02-27
[NEAT-801] New release process

doctrino:
# RISK REVIEWER FYI

The same process has already been approved for Toolkit: https://github.com/cognitedata/toolkit/pull/1487

# Description

After a discussion with the risk-reviewer team during the office hours 26/02, we are recommended to change our release process.

The changes are as follows:
* Changelog is no longer in the code base - only available in GitHub releases. This is to avoid unnecessary merge conflict of the changelog.
* A changelog entry should be part of the PR description. The PR description is now expected to follow the template shown below.
* The release process controls the versions, so the placeholder `0.0.0` is used in the main branch and only replaced just before release.

Note this depends on 
image

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added
- This will be ignored as I have checked skipped above.


nikokaoja (dismissed):

nikokaoja (approved):

.github/pull_request_template.md
```diff
@@ -0,0 +1,14 @@
+# Description
+
+Please describe the change you have made.
+
+## Bump
```
BugGambit, line 5: Even better is to use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) and semantic release

.github/workflows/release.yaml
```diff
@@ -72,43 +99,43 @@ jobs:
...
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+#
+#  release-to-f25e:
+#    runs-on: ubuntu-latest
+#    environment: main
+#    permissions:
+#      id-token: write
+#      contents: read
+#
```
BugGambit, line 109: Remove commented out code?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cognite-neat"
-version = "0.111.0"
+version = "0.0.0"
```
BugGambit, line 3: Can you explain we reset the version to 0.0.0? Isn't the package already published somewhere and we will have conflicts?

.github/workflows/release.yaml
```diff
@@ -25,39 +27,64 @@ jobs:
           poetry config virtualenvs.create false
           poetry install -E all
 
+      - name: Dump last commit message to file
```
BugGambit, line 30: What if the PR has multiple commits? We don't care about them?

FYI, it seems like we would our own semantic release process here.

doctrino: That is not possible, we require squash commit from PRs and the PR description becomes the commit message:
image

.github/pull_request_template.md
doctrino: That would cause the PR description to look strange. Note that we are using the PR descriptions, and not the individual commits to bump.

.github/workflows/release.yaml
```diff
@@ -25,39 +27,64 @@ jobs:
...
+            tag: ${{ env.version }}
+            bodyFile: last_changelog_entry.md
+
+#  release-to-docker-hub:
+#    runs-on: ubuntu-latest
+#    environment: CD
+#    steps:
+#      - name: Checkout repository
+#        uses: actions/checkout@v3
+#
```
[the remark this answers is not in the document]
doctrino: We are keeping it as we are planning to include it int the future, and I know we can lookup git history, but it is simply more convenient.

pyproject.toml
doctrino: The `0.0.0` is a placeholder. The CD process replaces it before release with the actual version. 
image

nikokaoja (approved):

BugGambit (approved):
Given this follows the same pattern outlined in the following approved PRs:
- https://github.com/cognitedata/pygen/pull/438
- https://github.com/cognitedata/toolkit/pull/1487

I will approve

PR #1022, opened 2025-02-28
[NEAT-804] ⛰ Skip running twice for hierarchical properties

doctrino:
# Description
We used to load CogniteAsset (and extensions) twice. This is not necessary, so removing the code for it.

**NOTE** Merge aftre new release process is in place #1019

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved
- The `neat.to.cdf.instances()` no longer loads CogniteAsset (and extensions) twice.

PR #1025, opened 2025-03-01
🤖 Allow Improved Header in Changelog.

doctrino:
# Description

Allow Improved header in changelog. In addition, no longer require a changelog entry when you select `skip` in the bump section.

**Note** Repeating the changelog from PR #1022 as that was not released due to the `Improved` header was not allowed. This is unlikely to happen in the future as a new check of the PR description will validate the bump and changelog entry. 

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved
- The `neat.to.cdf.instances()` no longer loads CogniteAsset (and extensions) twice.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,13 +33,13 @@ jobs:
         run: git describe --tags --abbrev=0 > last_version.txt
       - name: Bump Version
         run: python dev.py bump --verbose
-      - name: Create CHANGELOG entry
-        run: python dev.py changelog
       - id: version
         name: Read the new version
         # Read the version from the cognite/neat/_version.py file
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -33,13 +33,13 @@ jobs:
...
-      - name: Create CHANGELOG entry
-        run: python dev.py changelog
       - id: version
         name: Read the new version
         # Read the version from the cognite/neat/_version.py file
         run: echo "version=$(sed -n 's/^__version__ = "\(.*\)"/\1/p' cognite/neat/_version.py)" >> $GITHUB_ENV
-
+      - name: Create CHANGELOG entry
+        if: env.version != '0.0.0'
+        run: python dev.py changelog
```
doctrino, line 42: Don't try to create changelog when skip is selected.

nikokaoja (approved):

larshagencognite (approved):

PR #1028, opened 2025-03-01
🐛 Visualization multiple connections in same direction

doctrino:
# Description
### Before
image
#### After
image

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- The `neat.show.data_model()` now supports visualization of multiple connections in the same directions between two views/classes.


nikokaoja (approved):
Good catch !

finnag (approved):

finnag (approved):

PR #1033, opened 2025-03-04
[NEAT-810] 😦 No warning from reading core

doctrino:
# Description

### Before
image

### After
image

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed
- Reading the `CogniteCore` model, `neat.read.examples.core_data_model()`, no longer give warnings.


rsjr (approved):

nikokaoja (approved):
I see it can be annoying and confusing for users why we are raising raw filter warning, so approving

finnag (approved):

PR #1038, opened 2025-03-10
Metadata inferencing

doctrino:
# Description

Added explicit test of metadata inferencing in hunt for a bug. This is useful because we only have implicit testing of this functionality through snapshot testing. Also fixed a bug in the test data were metadata of a classic asset were something else that a string of string mapping.

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


rsjr (approved):

cemreyavuz (approved):
:unicorn:

PR #1039, opened 2025-03-10
🏋️‍♂️DMSExtractor use view external ID as rdf.type

doctrino:
# Description

In the uplift workflow, we expect he type that is read from a view to match the view externalId. If `node.type` is set, this assumptions can be broken. This PR uses the view external ID as the rdf type.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When reading instances using `neat.read.cdf.graph(...)`, neat now enforces the instances type to match the view external ID.


dbrattli (dismissed):
:unicorn:

cognite/neat/_graph/extractors/_dms.py
```diff
@@ -43,13 +43,15 @@ def __init__(
         unpack_json: bool = False,
         empty_values: Set[str] = DEFAULT_EMPTY_VALUES,
         str_to_ideal_type: bool = False,
+        ignore_node_type: bool = True,
```
nikokaoja, line 46: This should be config with literal values where one can indicate what should be used to set RDF type with options:

- Node type
- View external id
- ... 

This allows us for more flexibility in future

nikokaoja (changes requested):

cognite/neat/_graph/extractors/_dms.py
```diff
@@ -33,6 +33,10 @@ class DMSExtractor(BaseExtractor):
             considered as an empty value and skipped.
         str_to_ideal_type: If unpack_json is True, when unpacking JSON objects, if the value is a string, the extractor
             will try to convert it to the ideal type.
+        node_type: The prioritized order of the node type to use. The options are "view" and "type". "view_id"
+            means the externalId of the view used as type, while type is the node.type.
+        edge_type: The prioritized order of the edge type to use. The options are "view" and "type". "view_id"
+            means the externalId of the view used as type, while type is the edge.type.
```
doctrino, line 39:
```suggestion
        node_type: The prioritized order of the node type to use. The options are "view" and "type". "view"
            means the externalId of the view used as type, while type is the node.type.
        edge_type: The prioritized order of the edge type to use. The options are "view" and "type". "view"
            means the externalId of the view used as type, while type is the edge.type.
```

larshagencognite (approved):
risk review ok

PR #1040, opened 2025-03-10
🚝Timeout 408 issue

doctrino:
# Description

Neat is used as a migration tool. This requires downloading instances in the several hundred thousand count. A not uncommon response is
```
cognite.client.exceptions.CogniteAPIError: Graph query timed out. 
Reduce load or contention, or optimise your query. | code: 408 | 
X-Request-ID: bd0f77cd-ce4f-9889-8e7f-0b50b77780d6 | cluster: az-eastus-1
```

This PR implements a reduction in the number of instances neat asks for, with a gradual increase after a back-off when you get a 408. This is necessary to do this large downloads which again is necessary in the migration work. 

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- The `neat.read.cdf.graph(...)`, no longer stops if it gets a 408 response from CDF, but instead Neat try to recover.


cognite/neat/_client/_api/neat_instances.py
```diff
@@ -0,0 +1,67 @@
...
+        if filter_:
+            body["filter"] = filter_.dump() if isinstance(filter_, Filter) else filter_
+        if source:
+            body["sources"] = [{"source": source.dump(include_type=True, camel_case=True)}]
+        last_limit_change: float | None = None
+        while True:
+            try:
+                response = self._client.post(url=url, json=body)
+            except CogniteAPIError as e:
+                if e.code == 408 and body["limit"] > 1:
```
rsjr, line 51: Is it possible for us to have a generic retry decorator so we could reuse across neat?

I understand the instances endpoint is special, but the it would be nice for us to have both exp backoff and jitter as default. 

[Example](https://github.com/cognitedata/python-extractor-utils/blob/eaffda21c5da122bb3ca74809087c1dead6278e6/cognite/extractorutils/uploader/data_modeling.py#L69) on how it's done in the python extractor utils.

cognite/neat/_client/_api/neat_instances.py
```diff
@@ -0,0 +1,67 @@
...
+            response_body = response.json()
+            yield from (
+                Node.load(item) if item.get("instanceType") == "node" else Edge.load(item)
+                for item in response_body["items"]
+            )
+            next_cursor = response_body.get("nextCursor")
+            if next_cursor is None:
+                break
+            body["cursor"] = next_cursor
+            if last_limit_change is not None and (time.perf_counter() - last_limit_change) > 30.0:
```
rsjr, line 65: Why 30 seconds here? Any particular reason? I would say this might change depending how big is the payload / call.

cognite/neat/_client/_api/neat_instances.py
```diff
@@ -0,0 +1,67 @@
...
+                "direction": "ascending",
+            }
+        ]
+        url = f"/api/{self._client._API_VERSION}/projects/{self._client.config.project}/models/instances/list"
+        filter_ = self._client.data_modeling.instances._merge_space_into_filter(instance_type, space, filter_)
+        if filter_:
+            body["filter"] = filter_.dump() if isinstance(filter_, Filter) else filter_
+        if source:
+            body["sources"] = [{"source": source.dump(include_type=True, camel_case=True)}]
+        last_limit_change: float | None = None
```
rsjr, line 46: Nitpick: Why float or None? Could this be initialized as 0?

rsjr (commented):
We could to reuse what has been done in the extractor utils. https://github.com/cognitedata/python-extractor-utils/blob/eaffda21c5da122bb3ca74809087c1dead6278e6/cognite/extractorutils/uploader/data_modeling.py#L69

It's a standardized way to do retries, might be a good fit for NEAT as well :)

cognite/neat/_client/_api/neat_instances.py
doctrino: Added documentation to code.

rsjr (approved):

tests/tests_unit/graph/test_extractors/test_dms_graph_extractor.py
```diff
@@ -1,36 +1,37 @@
...
 from tests.utils import as_read_containers, as_read_instance, as_read_space
 
+_IS_CALLED = False
+
 
 def car_instances(
-    instance_type: Literal["node", "edge"] = "node", sources: Source | Sequence[Source] | None = None, **other_args
+    instance_type: Literal["node", "edge"] = "node", source: Source | None = None, **other_args
 ) -> Iterable[Instance]:
+    global _IS_CALLED
```
larshagencognite, line 21: This warrants a comment

doctrino: Comment included.

rsjr (approved):

PR #1042, opened 2025-03-12
[NEAT-823] 🧙‍♂️ Rename create to template

doctrino:
# Description

We realized that `create` is misleading, thus renaming this API to template as what we are creating is a template. For example `neat.template.solution_model(...)` reads more precisely what we are doing instead of `neat.create.solution_model(...)`. 

## Bump

- [ ] Patch
- [x] Minor
- [ ] Skip

### Changed

- [BREAKING] Renamed the `neat.create` to `neat.template`.


nikokaoja (approved):
This is great ! I fully agree !

What we should add in one of the next PR is that users can export fully empty templates for Info and DMS rules.

PR #1043, opened 2025-03-12
[NEAT-828] 👋 Class entities support version

doctrino:
# Description

We have not needed this functionality before now. Adding tests to verify that it works as intended. 

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/tests/tests_unit/rules/test_models/test_information_rules.py
+++ b/tests/tests_unit/rules/test_models/test_information_rules.py
@@ -507,7 +507,7 @@ class TestInformationClass:
     @pytest.mark.parametrize(
         "raw, class_, implements",
         [
-            pytest.param(
+            (
                 InformationInputClass(
                     class_="WindTurbine",
                     description="Power generating unite",
--- a/tests/tests_unit/rules/test_models/test_information_rules.py
+++ b/tests/tests_unit/rules/test_models/test_information_rules.py
@@ -501,3 +501,31 @@ class TestInformationProperty:
         prop = InformationProperty.model_validate(raw.dump(default_prefix="my_space"))
 
         assert prop.class_ == expected
+
+
+class TestInformationClass:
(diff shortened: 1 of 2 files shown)


rsjr (approved):

tests/tests_unit/rules/test_models/test_information_rules.py
```diff
@@ -482,3 +482,50 @@ def test_rdf_properties(self, raw: InformationInputProperty):
         prop = InformationProperty.model_validate(raw.dump(default_prefix="power"))
 
         assert isinstance(prop, InformationProperty)
+
+    @pytest.mark.parametrize(
+        "raw, expected",
+        [
+            pytest.param(
```
rbtcollins, line 489:
```suggestion
```

tests/tests_unit/rules/test_models/test_information_rules.py
```diff
@@ -482,3 +482,50 @@ def test_rdf_properties(self, raw: InformationInputProperty):
...
+        prop = InformationProperty.model_validate(raw.dump(default_prefix="my_space"))
+
+        assert prop.class_ == expected
+
+
+class TestInformationClass:
+    @pytest.mark.parametrize(
+        "raw, class_, implements",
+        [
+            pytest.param(
```
rbtcollins, line 510: I don't see any meta activities here - the pytest.param() call isn't needed (it isn't harmful either, but just some noise)

rbtcollins (commented):
:unicorn: though I do have a note :)

tests/tests_unit/rules/test_models/test_information_rules.py
doctrino: You are correct, removed it.

rsjr (approved):

rbtcollins:
:unicorn:

PR #1049, opened 2025-03-14
[NEAT-838] 😵‍💫 Default value in DMS Rules

doctrino:
# Description

When reading from excel a TRUE value is interpreted as `1.0` by pandas, this causes the validation to fail for `DMSProperty`

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Setting a default value other than text in the DMS model spreadsheet will no longer cause `neat.read.excel()` to raise a `TypeError`.


rsjr (commented):
Is it something maybe for us to investigate why pandas is parsing excel boolean to 1/0? I've seen this before, sometimes pandas tries to be smart.

This fix will do the trick, but I am just thinking if there is something else we could do in SpreadsheetRead class to mitigate this :)

doctrino:
> Is it something maybe for us to investigate why pandas is parsing excel boolean to 1/0? I've seen this before, sometimes pandas tries to be smart.
> 
> This fix will do the trick, but I am just thinking if there is something else we could do in SpreadsheetRead class to mitigate this :)

In this case, the default column for the DMS sheet supports lots of different types of default, float, int, bool, str. So in that case it will need something like this. So I am not sure if this is necessary here.

rsjr (approved):

cemreyavuz (approved):
risk review ok

PR #1058, opened 2025-03-20
[prep NEAT-852]🕵🏼DMSSchema from read model

doctrino:
# Description

Added an internal helper method to the `DMSSchema` that creates the schema from a read version of a data model. This is useful when we get model from a user and need this for debugging.

I used this one to hunt down the bug in #1059 

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


rsjr (approved):

larshagencognite (approved):
risk review ok

PR #1059, opened 2025-03-20
[NEAT-852] 𓀎 Raw filter filter with special character fix.

doctrino:
# Description

Caused by a naive use of regex. Replace it with startswith and endswith for more robust parsing. 

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Running `neat.read.cdf.data_model(...)` to fetch a data model no longer raises a `ValueError` if one of the views in the model contains a filter that uses a special characters such as `&`.

### DIFF
--- a/cognite/neat/_rules/models/entities/_wrapped.py
+++ b/cognite/neat/_rules/models/entities/_wrapped.py
@@ -42,7 +42,7 @@ class WrappedEntity(BaseModel, ABC):
         # raw filter case:
         if cls.__name__ == "RawFilter":
             if data.startswith("rawFilter(") and data.endswith(")"):
-                return {"filter": data.removeprefix("rawFilter(").removesuffix(")"), "inner": None}
+                return {"filter": data[10:-1], "inner": None}
             else:
                 raise ValueError(f"Cannot parse {cls.name} from {data}. Ill formatted raw filter.")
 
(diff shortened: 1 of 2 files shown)


cognite/neat/_rules/models/entities/_wrapped.py
```diff
@@ -193,7 +193,7 @@ class RawFilter(DMSFilter):
     filter: str
     inner: None = None  # type: ignore[assignment]
 
-    def as_dms_filter(self) -> dm.Filter:  # type: ignore[override]
+    def as_dms_filter(self, default: Any | None = None) -> dm.Filter:  # type: ignore[override]
```
doctrino, line 196: Unrelated to the bugfix.. This was just an inconsistency, the base class `DMSFilter.as_dms_filter(...)` defines this method with this parameter.

cognite/neat/_rules/models/entities/_wrapped.py
```diff
@@ -42,8 +41,8 @@ def _parse(cls, data: str) -> dict:
 
         # raw filter case:
         if cls.__name__ == "RawFilter":
-            if match := re.search(r"rawFilter\(([\s\S]*?)\)", data):
-                return {"filter": match.group(1), "inner": None}
+            if data.startswith("rawFilter(") and data.endswith(")"):
+                return {"filter": data[10:-1], "inner": None}
```
rsjr, line 45: this slice here is very specific. Any possiblity of having out-of-bounds errors here when doing this? Looks a bit dangerous, but I lack a bit on the context here

cognite/neat/_rules/models/entities/_wrapped.py
rsjr: Not for now maybe, but we should try to avoid "type: ignore".

larshagencognite (approved):
risk review ok

PR #1063, opened 2025-03-23
fix: Capture correct information for IMF-types.

bhtvedt:
An IMF_base_model has been introduced in CDF to capture core information about IMF-types as well as provide the link to the IMF_Standard_model.

This commit updates the classes and properties extracted from the IMF- types, so that it is in accordance with the IMF_base_model. It also include a mapping from:
    imf:Terminal -> imf:base:Terminal(version=v1)
    imf:Block -> imf:base:Block(version=v1)

Minor additional changes are:
- Custom strategy for making identifiers CDF compliant
- IMF related metadata

# Description

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

The `IMF Importer` produces cleaner and more accurate IMF instance models

### DIFF
--- a/cognite/neat/_rules/importers/_base.py
+++ b/cognite/neat/_rules/importers/_base.py
@@ -24,27 +24,32 @@ class BaseImporter(ABC, Generic[T_InputRules]):
         raise NotImplementedError()
 
     def _default_metadata(self) -> dict[str, Any]:
-        creator = "UNKNOWN"
-        with suppress(KeyError, ImportError):
-            import getpass
-
-            creator = getpass.getuser()
-
         return {
             "schema": "partial",
             "space": "neat",
             "external_id": "NeatImportedDataModel",
             "version": "0.1.0",
             "name": "Neat Imported Data Model",
-            "created": datetime.now().replace(microsecond=0).isoformat(),
-            "updated": datetime.now().replace(microsecond=0).isoformat(),
-            "creator": creator,
+            "created": self._get_current_time(),
+            "updated": self._get_current_time(),
+            "creator": self._get_username(),
             "description": f"Imported using {type(self).__name__}",
         }
 
     @classmethod
     def _repr_html_(cls) -> str:
         return class_html_doc(cls)
+
+    @classmethod
+    def _get_username(cls) -> str:
+        with suppress(KeyError, ImportError):
+            import getpass
+            return getpass.getuser()
+        return "UNKNOWN"
+
+    @classmethod
+    def _get_current_time(cls) -> str:
+        return datetime.now().replace(microsecond=0).isoformat()
 
     @property
     def agent(self) -> "ProvenanceAgent":
--- a/cognite/neat/_rules/importers/_rdf/_base.py
+++ b/cognite/neat/_rules/importers/_rdf/_base.py
@@ -123,8 +123,8 @@ class BaseRDFImporter(BaseImporter[InformationInputRules]):
             raise MultiValueError(self.issue_list.errors)
 
         rules_dict = self._to_rules_components()
-
         rules = InformationInputRules.load(rules_dict)
+
         self.issue_list.trigger_warnings()
         return ReadRules(rules, {})
 
@@ -145,12 +145,13 @@ class BaseRDFImporter(BaseImporter[InformationInputRules]):
     @property
     def _metadata(self) -> dict:
         return {
+            "schema": "partial",
             "role": RoleTypes.information,
             "space": self.data_model_id.space,
             "external_id": self.data_model_id.external_id,
             "version": self.data_model_id.version,
-            "created": datetime.now().replace(microsecond=0),
-            "updated": datetime.now().replace(microsecond=0),
+            "created": self._get_current_time(),
+            "updated": self._get_current_time(),
             "name": None,
             "description": f"Data model imported using {type(self).__name__}",
             "creator": "Neat",
--- a/cognite/neat/_rules/importers/_rdf/_imf2rules.py
+++ b/cognite/neat/_rules/importers/_rdf/_imf2rules.py
@@ -1,67 +1,60 @@
 """This module performs importing of various formats to one of serializations for which
 there are loaders to TransformationRules pydantic class."""
+import re
+import copy
+from uuid import UUID
+from pathlib import Path
+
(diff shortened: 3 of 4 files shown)


cognite/neat/_rules/importers/_base.py
```diff
@@ -24,28 +24,34 @@ def to_rules(self) -> ReadRules[T_InputRules]:
...
             "schema": "partial",
             "space": "neat",
             "external_id": "NeatImportedDataModel",
             "version": "0.1.0",
             "name": "Neat Imported Data Model",
-            "created": datetime.now().replace(microsecond=0).isoformat(),
-            "updated": datetime.now().replace(microsecond=0).isoformat(),
-            "creator": creator,
+            "created": self._get_current_time(),
+            "updated": self._get_current_time(),
```
doctrino, line 34: Lets call this one above and store it such that we are sure that `created` and `updated` are equal for new models
```python
        created_time = self._current_time()
        return {
            ...
            "created": created_time,
            "updated": created_time,
          ...
        }
```

cognite/neat/_rules/importers/_rdf/_base.py
```diff
@@ -145,12 +143,13 @@ def _add_uri_namespace_to_prefixes(cls, URI: URIRef, prefixes: dict[str, Namespa
...
         return {
+            "schema": "partial",
             "role": RoleTypes.information,
             "space": self.data_model_id.space,
             "external_id": self.data_model_id.external_id,
             "version": self.data_model_id.version,
-            "created": datetime.now().replace(microsecond=0),
-            "updated": datetime.now().replace(microsecond=0),
+            "created": self._get_current_time(),
+            "updated": self._get_current_time(),
```
doctrino, line 152: Same here

cognite/neat/_rules/importers/_rdf/_imf2rules.py
```diff
@@ -82,14 +81,76 @@ def _to_rules_components(
         self,
     ) -> dict:
         classes, issue_list = parse_classes(self.graph, CLASSES_QUERY, self.language, self.issue_list)
+        mapped_classes = self._map_to_base_model(classes)
+        compliant_classes = self._make_compliant(mapped_classes, ["class_"])
```
doctrino, line 85: We try to avoid this pattern of doing logic in the imports, instead have that logic separately in a transformer to avoid repeating it in each transformer. 

Can this be done here as well? 

If this is something you always should do when you read IMF, then it is ok to call this transformer automatically in the NeatSession when reading IMF.

cognite/neat/_rules/importers/_rdf/_imf2rules.py
```diff
@@ -82,14 +81,76 @@ def _to_rules_components(
...
+
         self.issue_list = issue_list
+
         properties, issue_list = parse_properties(self.graph, PROPERTIES_QUERY, self.language, self.issue_list)
+        compliant_properties = self._make_compliant(properties, ["class_", "property_"])
```
doctrino, line 90: Same here

tests/tests_unit/rules/test_importers/test_imf_importer.py
```diff
@@ -12,7 +12,7 @@ def test_imf_importer():
...
     regex_violations = [issue for issue in issues if isinstance(issue, ResourceRegexViolationWarning)]
 
-    assert len(rules.classes) == 63
-    assert len(rules.properties) == 62
-    assert len(issues) == 207
-    assert len(regex_violations) == 129
+    assert len(rules.classes) == 7
+    assert len(rules.properties) == 75
+    assert len(issues) == 2
+    assert len(regex_violations) == 0
```
doctrino, line 18: Could you also extend this to also check which classes you expect, as well some samples for properties you expect?

cognite/neat/_rules/importers/_rdf/_imf2rules.py
```diff
@@ -82,14 +81,76 @@ def _to_rules_components(
...
+        # removing any double underscores that could occur
+        return re.sub(r"[^a-zA-Z0-9]+", "_", entity)
+
+    @classmethod
+    def _is_valid_uuid(cls, uuid_to_test: str, version: int = 4) -> bool:
+        try:
+            uuid_obj = UUID(uuid_to_test, version=version)
+            return str(uuid_obj) == uuid_to_test
+        except ValueError:
+            return False
```
doctrino, line 156: This is similar to logic we have in the `ToCompliantEntities` transformer. Can you use that instead?

doctrino (commented):
A few issues. Otherwise good fixes.

cognite/neat/_rules/importers/_rdf/_imf2rules.py
bhtvedt: Yes, I have 
- updated `ToCompliantEntities` add a prefix if the identifier is a UUID regardless of whether or not is starts with a number. This ensures that all the UUIDs are treated the same. Further more, it is now possible to supply a custom prefix and suffix.
- added a new `IMFRuleMapper` utilizing the existing `RuleMapper`. At the moment it is only implements that are mapped, but I believe that there will be more such mappings. Now the two mappings are hardcoded, but in the future they will be read from file.

The two bullet points ensure that all non-import related logic is moved to post-processing steps.

cognite/neat/_rules/importers/_rdf/_imf2rules.py
bhtvedt: See above

tests/tests_unit/rules/test_importers/test_imf_importer.py
bhtvedt: Done, the tests have been expanded.

cognite/neat/_rules/importers/_rdf/_imf2rules.py
bhtvedt: This section is now removed. The reason for the duplication was simply me not wanting to add IMF custom logic to your generic components. But I believe this now has been sorted in a satisfactory way.

bhtvedt:
> A few issues. Otherwise good fixes.

Thanks for the feedback. A few pointers on how you prefer to structure the code helps a lot. 👊

PR #1066, opened 2025-03-26
NEAT-858 Missing labels in ontology

nikokaoja:
# Description

When creating ontology out of DMS rules we were not properly creating triple that holds display `name` of  `View` and `property`. The issue was on one part in conversion from DMS rules to Information rules, and then proper handling of creation of triple that hold `rdfs:label` and `dcterm:title`.

## Bump

- [X] Patch
- [ ] Minor
- [ ] Skip

### Fixed
- Conversion of `name` from DMS rules to Information rules
- Conversion of `name` from Information rules to Ontology triple for classes and properties

### Added
- Creation of triple holding information on title which generated by combining external id and name


cognite/neat/_rules/transformers/_converters.py
```diff
@@ -1766,6 +1767,7 @@ def as_information_rules(
                 # Removing version
                 class_=ClassEntity(suffix=property_.view.suffix, prefix=property_.view.prefix),
                 property_=property_.view_property,
+                name=property_.name,
```
doctrino, line 1770: Good catch

doctrino (approved):

psalaberria002:
The PR title should say what the PR is doing. Missing labels in ontology seems the be the reason why you are creating the PR.

cognite/neat/_rules/exporters/_rules2ontology.py
```diff
@@ -289,6 +289,19 @@ def label_triples(self) -> list[tuple]:
         else:  # If comment is None, return empty list
             return []
 
+    @property
+    def title_triples(self) -> list[tuple]:
```
psalaberria002, line 293: What are the types for each variable in the tuple? `list[tuple[str, str, str]]` for example?

psalaberria002 (approved):
:unicorn: 

More narrowed type hints may be useful here. Relying on the function name to deduced how many elements are in those tuples doesn't seem great.

cognite/neat/_rules/exporters/_rules2ontology.py
nikokaoja: @psalaberria002 agree! I will make a PR to make type hints more detailed

PR #1070, opened 2025-03-28
Bump to latest Pydantic version

rsjr:
# Description

Bump to Pydantic 2.11. 
For reviewers: The small change in the _contextmanager warning handler is to skip loging Pydantic Deprecation warning to end users. 

We are going to fix those along the way for sure, but after bumping to stable Pydantic 2.11 warnings were getting sent to  NEAT endusers.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Dependency update (pydantic) 
- Skip PydanticDeprecationWarning to the end users


psalaberria002 (approved):
:unicorn:

PR #1071, opened 2025-03-31
[NEAT-867] 🧹 Remove unused fixtures/tests and outdated example rules

doctrino:
# Description

Removed unused fixtures and skipped tests that are based on a year old models that are no longer valid, thus these tests no longer serves a purpose and gives a false sense of coverage.

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


rsjr (approved):
🚀

psalaberria002 (approved):
:unicorn:

PR #1072, opened 2025-03-31
Ci: update coverage threshold

doctrino:
# Description

We introduced the coverage test a year ago, and then I set a low threshold as we had a lot of uncovered code. Suggest we now increase it to 80% which is a good target for coverage. 

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


rsjr (approved):
Nice

PR #1076, opened 2025-04-01
[NEAT-873] 🪓 Split write and read for queries

doctrino:
# Description

**Context**: NeatStore is the graph storage used by neat. That store has an API `.queries` that has methods for performing standardized queries against the store. This PR splits this into `queries.select...` and `queries.update...`. 
The motivation for this is to add extra checks before the user is allowed to do modifications to the store. This PR is a first step.

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite/neat/_graph/queries/_write.py
```diff
@@ -0,0 +1,33 @@
+from rdflib import Dataset, URIRef
+
+from cognite.neat._utils.rdf_ import remove_instance_ids_in_batch
+
+from ._base import BaseQuery
+from ._read import ReadQueries
+
+
+class WriteQueries(BaseQuery):
```
nikokaoja, line 9:
```suggestion
class WriteQueries(BaseQuery):
    """This class holds a set of SPARQL queries which are updating triples in the knowledge graph.
    The update queries are executed against update endpoint, and typically start with UPDATE statement
    """
```

cognite/neat/_graph/queries/_read.py
```diff
@@ -0,0 +1,436 @@
...
+from rdflib.query import ResultRow
+
+from cognite.neat._constants import NEAT
+from cognite.neat._shared import InstanceType
+from cognite.neat._utils.rdf_ import remove_namespace_from_uri
+
+from ._base import BaseQuery
+
+
+class ReadQueries(BaseQuery):
```
nikokaoja, line 17:
```suggestion
class ReadQueries(BaseQuery):
    """This class holds a set of SPARQL queries which are reading triples from the knowledge graph.
    The read queries are executed against query endpoint, and typically start with SELECT statement.
    """
```

thorkildcognite:
And thanks for doing the refactor in one go, and not get tempted to rewrite them all in the process. Made it a lot easier to compare and see that the logic had not changed beyond getting moved.

PR #1077, opened 2025-04-01
[NEAT-874] 🔨 Interface for state

doctrino:
# Description

Todo

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite/neat/_state/_base.py
```diff
@@ -0,0 +1,90 @@
...
+    def export_to_file(self, exporter: BaseExporter, path: Path) -> None:
+        """Export the rules to a file."""
+        raise NotImplementedError
+
+    def export_to_cdf(self, exporter: CDFExporter, client: NeatClient, dry_run: bool) -> UploadResultList:
+        """Export the rules to CDF."""
+        raise NotImplementedError
+
+    def load(self, loader: BaseLoader) -> UploadResultList:
+        """Load the instances into CDF."""
```
nikokaoja, line 89: should this be one method controlled with arguments, they are all doing the same thing "exporting" from session to somewhere. 

def export(resource : Literal["instances", "data_model"], exporter: BaserExporter | CDFExporter,  client: NeatClient , dry_run: bool | None = None)

cognite/neat/_state/_state.py
```diff
@@ -0,0 +1,65 @@
...
+from cognite.neat._graph.extractors import BaseExtractor
+from cognite.neat._graph.transformers import BaseTransformer
+from cognite.neat._rules.importers import BaseImporter
+from cognite.neat._rules.transformers import InformationToDMS, VerifiedRulesTransformer
+from cognite.neat._store import NeatGraphStore, NeatRulesStore
+
+from ._types import Action
+
+
+class State(ABC):
```
nikokaoja, line 12: What is the difference between State and NeatState? Is NeatState inherenting from State?

cognite/neat/_state/_state.py
```diff
@@ -0,0 +1,65 @@
...
+    @property
+    def display_name(self) -> str:
+        return type(self).__name__.removesuffix("State")
+
+    @abstractmethod
+    def is_valid_transition(self, action: Action) -> bool:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def next_state(self, action: Action) -> "State":
```
nikokaoja, line 26: Next state is telling you what is possible next state or ?

cognite/neat/_state/_state.py
```diff
@@ -0,0 +1,65 @@
...
+        raise NotImplementedError()
+
```
nikokaoja, line 28: Would be helpful to have state and flow digram depicting this, as this is to become heart of neat

cognite/neat/_state/_state.py
```diff
@@ -0,0 +1,65 @@
...
+        elif isinstance(action, BaseImporter):
+            return ConceptualState(self._rule_store, self._graph_store)
+        raise NotImplementedError()
+
+
+class InstancesState(State):
+    def is_valid_transition(self, action: Action) -> bool:
+        return isinstance(action, BaseTransformer)
+
+    def next_state(self, action: Action) -> "State":
```
nikokaoja, line 46: would next_state be again instance state as you are able to do manipulation on instances?

nikokaoja (commented):
Would help with flow and state diagram

cognite/neat/_state/_state.py
doctrino: No, it is transitioning to the next state.

cognite/neat/_state/_base.py
```diff
@@ -0,0 +1,90 @@
...
+from cognite.neat._utils.upload import UploadResultList
+
+from ._state import EmptyState, State
+from ._types import Action
+from .exception import InvalidStateTransition
+
+
+# Todo: This class is in progress and not currently used. Through a series of PRs, it will replace the
+#  SessionState class as well as move the logic from the _session into this _state module.
+class NeatState:
```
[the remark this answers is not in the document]
doctrino: No, `State` is internal and should not be exposed to the rest of the code base. This is the external facing state. Renamed to make it clearer.

cognite/neat/_state/_state.py
doctrino: Will do, but that will have to be in a separate PR, I have captured it in a task. The reason is to keep these refactoring PRs small and the goal of this PR is the mechanism of the state machine not the logic of the transitions.

cognite/neat/_state/_base.py
doctrino: I don't think so, it will lead to lots of overloads as they return slightly different results. Shared for all these methods is that they are not changing the state.

cognite/neat/_state/_state.py
doctrino: See answer below

cognite/neat/_state/_state.py
doctrino: Not necessarily. From an instance state you can move to the conceptual state if you get a schema from the transformer. The reason why the logic is missing here, is that we are missing some logic as well as some of that logic is currently in the NeatSession and needs to be moved out.

cognite/neat/_state/_base.py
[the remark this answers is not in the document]
doctrino: An alternative is to call this `NeatManager` or something like that, but I think that is a too generic name that does not tell you much as a developer.

cognite/neat/_state/_base.py
nikokaoja: `export` is returning `T_Export` that can be wide variety of outputs:
- DMSSchema
- Workbook
- Graph
- str

therefore, you already with `export` are having export a spectrum of output types, though it is hidden under `T_Export` type.

Therefore, you already have what you mentioned in your reply a slightly different results for `export` method.

So either we are explicit and have export method per output type, or we have single export method but then control its behaviour with arg.

Currently it is a mixed

nikokaoja (approved):

nikokaoja (changes requested):
rename action to activity to be in sync with provenance

cognite/neat/_state/_base.py
```diff
@@ -0,0 +1,90 @@
...
+        self._rule_store = NeatRulesStore()
+        self._graph_store = NeatGraphStore.from_memory_store()
+        self._state: InternalState = EmptyState(self._rule_store, self._graph_store)
+
+    @property
+    def status(self) -> str:
+        """Returns the display name of the current state."""
+        return self._state.display_name
+
+    def change(self, action: Action) -> IssueList:
```
nikokaoja, line 40: This is change in terms of what we define as change in Provenance, therefore, your argument is not action, but Agent (extractor, loader, transformer) that is configured to perform certain Activity.

PR #1079, opened 2025-04-01
Enable dummy property per user defined concept

nikokaoja:
# Description

Enable adding dummy property for every user-defined concept when creating extension template.
By default dummy property is formed as `idOfConceptGUID` , and it set to be string, none mandatory.
Dummy properties help in avoiding to set filters in DMS.

## Bump

- [ ] Patch
- [x] Minor
- [ ] Skip

### Added
- Support for dummy properties in `template.extension()`


cognite/neat/_session/_template.py
```diff
@@ -206,6 +211,8 @@ def extension(self, io: Any, output: str | Path | None = None) -> IssueList:
             io: The input spreadsheet.
             output: The output spreadsheet. If None, the output will be the same
                 as the input with `_extension` added to the name.
+            dummy_property: The dummy property to use as placeholder for user-defined properties
+                            for each user-defined concept.
```
doctrino, line 215: Is it worth mentioning filtering as well? Something like
```suggestion
            dummy_property: The dummy property to use as placeholder for user-defined properties
                            for each user-defined concept. When converting a data model, it is recommended to have at least
                            one property for each concept. This ensures that you follow that recommendation.
```

doctrino (dismissed):

psalaberria002 (commented):
Please add a test which uses this field.

tests/tests_integration/test_session/test_graph_flow/test_create_extension_template.yml
```diff
@@ -263,6 +263,12 @@ properties:
   neatId: http://purl.org/cognite/neat/data-model/verified/logical/my_space/MyModel/v1/PowerGeneratingUnit/pathLastUpdatedTime
   property_: pathLastUpdatedTime
   value_type: dateTime
```
nikokaoja, line 265: Regression test data are accordingly updated containing expected new dummy properties that end with GUID

nikokaoja: @psalaberria002 ref your comment ^

nikokaoja:
> psalaberria002

Test is already in place, you can see that the regression test data is updated:
tests/tests_integration/test_session/test_graph_flow/test_create_extension_template.yml

I've tagged you at specific parts of test data where you can see the outcome of new feature

rsjr (approved):

PR #1080, opened 2025-04-02
[NEAT-881]🫣Allow multi value hack for reverse direct relations.

doctrino:
# Description

DMS does not allow EXPLICIT multi value direct relation properties. However, if you set the `source` of a direct relation property to None, and have multiple reverse direct relations using this property, Search will give you the behavior of a multi value direct relation. Before this PR, this type of model could not be read into Neat as Neat will give you a `ReversedConnectionNotFeasibleError` stopping the user from reading a published model into CDF.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- The `neat.read.cdf.data_model(...)` no longer gives an error if the model contains reverse direct relation properties through a direct relation property that does not have source set.


rsjr (approved):

rsjr (approved):

larshagencognite (approved):
🦄

PR #1083, opened 2025-04-03
[NEAT-886] 👮 NeatGraphStore.transform catch issues

doctrino:
# Description

This is part of a refactoring to NeatState, #1077. This modifies the `.transform()` method of  `NeatGraphStore` to return an `IssueList`. The motivation for this is that this should comply with the other action methods  `NeatRuleStore.import_rules/transform()` and `NeatGraphStore.write` that already return an `IssueList`. In addition, it now returns an error if the graph transformation failed. This is important as a failed transformation execution should not lead to a state change. 

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite/neat/_store/_graph_store.py
```diff
@@ -319,45 +320,43 @@ def _add_triples(
         """
         add_triples_in_batch(self.graph(named_graph), triples, batch_size)
 
-    def transform(self, transformer: Transformers, named_graph: URIRef | None = None) -> None:
+    def transform(self, transformer: Transformers, named_graph: URIRef | None = None) -> IssueList:
```
doctrino, line 323: Note that all the previous warnings are now considered errors. This is because all the conditionals leads the transformations to be not executed, thus not successful and should thus not lead to a state change.

doctrino:
> Just to confirm: Although related to the upcoming refactor, this is a good change for the current scenario, right?
> 
> The fact it was returning None was a problem, since the beginning, looks like. Question is: Is it a breaking change for the end users which are currently using neat? No right?

Yes this is a good change independent of refactor. No it is not a breaking change.

rsjr (approved):

psalaberria002 (approved):
:unicorn:

PR #1084, opened 2025-04-03
[NEAT-887] 🤖 Introduce interfaces needed for NeatState

doctrino:
# Description

This is part of a refactoring to NeatState, #1077. 

It introduced interfaces needed for NeatState. The interface introduced are:
* `Action` - anything that can change the State.
* `PhysicalRulesTransformer` and `ConceptualRulesTransformer` - each can only be applied in the Physical and Conceptual state, respecively, and does not lead to state change. 

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite/neat/_graph/extractors/_base.py
```diff
@@ -6,14 +6,14 @@
...
-from cognite.neat._shared import Triple
+from cognite.neat._shared import Action, Triple
 from cognite.neat._utils.auxiliary import class_html_doc
 
 if TYPE_CHECKING:
     from cognite.neat._store._provenance import Agent as ProvenanceAgent
 
 
-class BaseExtractor:
+class BaseExtractor(Action):
```
nikokaoja, line 16: Why would BaseExtractor take class Method? All I can see is that `Action` is an abstract class? 
Also why making new conception `Action`? Can we use concept `Activity` from provenance?

Extractor/Transformer/Loader are Agents doing certain Activity

cognite/neat/_rules/importers/_base.py
```diff
@@ -7,13 +7,14 @@
 
 from cognite.neat._constants import DEFAULT_NAMESPACE
 from cognite.neat._rules._shared import ReadRules, T_InputRules
+from cognite.neat._shared import Action
```
nikokaoja, line 10: Is Action actually Activity from provenance ? An Agent (importer) is performing certain Activity (action).
Would it make sense to

nikokaoja (commented):
I am not totaly sure what `Action` is suppose to do

cognite/neat/_graph/transformers/_base.py
```diff
@@ -25,25 +25,33 @@ class RowTransformationOutput:
     instances_modified_count: int = 0
 
 
-class BaseTransformer(ABC):
-    description: str
+class BaseTransformer(Action):
+    _description: str
```
nikokaoja, line 29: why private?

nikokaoja (commented):
I am not totaly sure what `Action` is suppose to do

cognite/neat/_graph/extractors/_base.py
doctrino: It is in the PR description, the intention of the Action is anything that may change the state.

cognite/neat/_rules/importers/_base.py
doctrino: Hmm, that is up for discussion. An action is something that changes the state of Neat, while activity is anything you can perform. For example, exporters and loaders are activities that can be performed, but they are not actions as they will never lead to state change.

nikokaoja: Not correct, exporters and loaders are Agents, that when employed are executing certain Activity , therefore action == activity

cognite/neat/_graph/extractors/_base.py
nikokaoja: I understood in the referenced PR that `Action` is type, but in this PR it is converted to abstract class, which is then used by any type of what in provenance we call "Agents" (i.e. extractors, transformers, etc.).

PR #1095, opened 2025-04-10
fix inference bug

nikokaoja:
# Description

Fix bug with partial inference which requires Cognite client, previously missing

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Data model inference bug due to missing Cognite client


tests/tests_integration/test_session/test_graph_flow.py
```diff
@@ -144,7 +144,8 @@ def test_uplift_workflow_to_cdf(self, cognite_client: CogniteClient) -> None:
             unpack_json=True,
             str_to_ideal_type=True,
         )
-        neat.infer()
+        issues = neat.infer()
+        assert not issues.has_errors
```
nikokaoja, line 148: This test is affected with the change in the code. Previously we did not check if inference is completed without issues (which were there before!), now this assertation is making sure things are correct.

rsjr (approved):

dmivankov:
:unicorn:

PR #1096, opened 2025-04-14
NEAT-899 Remove  template. solution_model() 

nikokaoja:
# Description

Removing `template.solution_model()` as it is not robust to be public method

## Bump

- [X] Patch
- [ ] Minor
- [ ] Skip

###  Removed

- Removed `template.solution_model()` as it is not robust to be public method


rsjr (approved):

thorkildcognite:
risk review ok

PR #1098, opened 2025-04-15
Remove non-functional component selection

nikokaoja:
# Description

Sub-selection of schema components to be published to CDF does not work as expected in `neat.to.cdf.data_model()`. Accordingly this capability for time being will be removed from public method, since users can control what effect the method has via `existing` argument.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Removed

- Ability to fine-grain which component of schema is to be published to CDF

### DIFF
--- a/cognite/neat/_session/_to.py
+++ b/cognite/neat/_session/_to.py
@@ -1,6 +1,5 @@
 import warnings
 import zipfile
-from collections.abc import Collection
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


rsjr (approved):

simenl (commented):
🦄

simenl (approved):
🦄

PR #1100, opened 2025-04-23
Fix bug of reading In filter from CDF to NEAT

nikokaoja:
# Description

`IN` filter was read as `NodeType` by default from CDF, and reading did shallow check of the actual filter content. As result, `IN` filters, which for example performed checks if a view property value is in the range of values, was mistakenly read as `NodeType` instead of `RawFilter`. This led to errors in importing data model from CDF.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Fixed bug with reading RawFilters from CDF, especially `IN` filters which are not `NodeType` filters


rsjr (dismissed):
🥇

rsjr (approved):

cognite/neat/_rules/models/entities/_wrapped.py
```diff
@@ -107,12 +107,20 @@ def from_dms_filter(cls, filter: dm.Filter) -> "DMSFilter":
...
+            and (space := value.get("space"))
+            and (external_id := value.get("externalId"))
+        ):
+            return NodeTypeFilter(inner=[DMSNodeEntity(space=space, externalId=external_id)])
+        elif (
+            (body := dumped.get(dm.filters.In._filter_name))
+            and (values := body.get("values"))
+            and all(
+                [
+                    (True if isinstance(entry, dict) and "space" in entry and "externalId" in entry else False)
```
larshagencognite, line 119: not blocking, but why do you need `True if ... else False` here?

larshagencognite (approved):
🦄

PR #1145, opened 2025-04-28
bump experimental version

rsjr:
# Description

## Bump

- [ ] Patch
- [x] Minor
- [ ] Skip

### Added

- Release test experimental version

### DIFF
--- a/.github/workflows/release-experimental.yaml
+++ b/.github/workflows/release-experimental.yaml
@@ -75,21 +75,21 @@ jobs:
           twine check dist/*
           unzip -p dist/*.whl METADATA | grep -E '^(Name|Version):'
 
-      # - name: Release to PyPI
-      #   if: env.version != '0.0.0'
-      #   env:
-      #     TWINE_USERNAME: __token__
-      #     TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
-      #   run: uv run twine upload --skip-existing --verbose dist/*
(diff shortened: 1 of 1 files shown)


.github/workflows/release-experimental.yaml
```diff
@@ -75,21 +75,21 @@ jobs:
...
-      #       prerelease: false
-      #       draft: false
-      #       tag: ${{ EXPERIMENTAL_TAG }}
-      #       bodyFile: last_changelog_entry.md
+      - name: Release to PyPI
+        if: env.version != '0.0.0'
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
+        run: uv run twine upload --skip-existing --verbose dist/*
```
doctrino, line 83: Are we not releasing this to jfrog?

rsjr: That would be the best, but users won't be able to run this in CDF notebooks right? If it's in jfrog you must have access to the artifactory. 

If we want to keep experimental only for local usage, then I can change this to publish to jfrog. Wdyt?

doctrino: My I misunderstood something, but if we release to pypi then we need risk-reviewers again. Which is at least 50% of the reason for why we are doing this in the first place.

doctrino (approved):

PR #1155, opened 2025-05-02
Experiment clustering

doctrino:
# Description

Base branch for experimenting with clustering.

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


doctrino:
experiment concluded. Naive clustering does not work. You need to use human ingenuity to find patterns in the data.

PR #1160, opened 2025-05-02
[prep ] Refactor inference

doctrino:
# Description

Currently in the `neat.infer()` there is a standardization of property names that is done to merge properties that have a slightly different spelling. 

This causes problem were I want to use the inference without doing this standardization (working with Japanese characters that cannot be standardized to `[a-zA-Z0-9]`.

I fix this by moving out the standardization logic to a separate transformer that I call in the `neat.infer()` method.

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite/neat/_rules/importers/_rdf/_inference2rules.py
```diff
@@ -436,7 +436,7 @@ def _create_classes_properties(
                 shared_property_uris = set()
             shared_properties: dict[URIRef, list[_ReadProperties]] = defaultdict(list)
             for type_uri, properties_by_property_uri in properties_by_class_by_property.items():
-                class_suffix = remove_namespace_from_uri(type_uri)
+                class_suffix = urllib.parse.unquote(remove_namespace_from_uri(type_uri))
```
doctrino, line 439: Fix

cognite/neat/_rules/importers/_rdf/_inference2rules.py
```diff
@@ -457,35 +457,18 @@ def _create_classes_properties(
...
-                        if not isinstance(existing_prop.instance_source, list):
-                            existing_prop.instance_source = (
-                                [existing_prop.instance_source] if existing_prop.instance_source else []
-                            )
-                        existing_prop.instance_source.append(property_uri)
-                        continue
-                    else:
-                        properties_by_id[property_id_standardized] = self._create_property(
-                            read_properties, class_suffix, property_uri, property_id, prefixes
-                        )
```
doctrino, line 471: No longer necessary

cognite/neat/_rules/transformers/_converters.py
```diff
@@ -168,7 +171,6 @@ def _reserved_property_words(self) -> set[str]:
     def _fix_cls_suffix(self, suffix: str) -> str:
         if suffix in self._reserved_class_words:
             return f"My{suffix}"
-        suffix = urllib.parse.unquote(suffix)
```
doctrino, line 171: This is done in the `NamingStandardization`, so no need to do it here

PR #1163, opened 2025-05-07
[NEAT-957] 🏴‍☠️ Statistics endpoint

doctrino:
# Description

Context: We recently experienced an incident where Neat exceeded the instance limit for a project, causing InField in production to stop accepting new writes. This is the first part of the mitigation process, which involves implementing support for checking the instance limit using the NeatClient.

The reason this functionality is not included in the cognite-sdk is that the endpoint is not yet ready and should not be exposed to end users through the SDK. However, in this pull request, it is not directly exposed to the end user. In the next pull request, I will implement a check to be performed before loading instances into CDF.

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


rsjr (approved):
This is good stuff. Later I assume we will have support in the Python SDK, so we just need to remember to update the "custom" calls you made. ;)

cognite/neat/core/_client/_api/statistics.py
```diff
@@ -0,0 +1,91 @@
...
+from cognite.client.utils.useful_types import SequenceNotStr
+
+from cognite.neat.core._client.data_classes.statistics import (
+    ProjectStatsAndLimits,
+    SpaceInstanceCounts,
+    SpaceInstanceCountsList,
+)
+
+
+class StatisticsAPI(APIClient):
```
rsjr, line 17: nitpick (not blocker): For other people reviewing this code, the name of the class might give the impression that this is an statiticsAPI inside neat, but this is actually just consuming from `/models/statistics` endpoint from data modelling. Maybe some comments or even changing the name of the class might help

rsjr (commented):
Padawan risk reviewer: LGTM. This feature adds no risk, as it's not published to NEAT end users. Tests are OK. Waiting for mentor feedback

cognite/neat/core/_client/data_classes/statistics.py
```diff
@@ -0,0 +1,125 @@
...
+            "views": self.views.dump(camel_case=camel_case),
+            "dataModels" if camel_case else "data_models": self.data_models.dump(camel_case=camel_case),
+            "containerProperties" if camel_case else "container_properties": self.container_properties.dump(
+                camel_case=camel_case
+            ),
+            "instances": self.instances.dump(camel_case=camel_case),
+            "concurrentReadLimit" if camel_case else "concurrent_read_limit": self.concurrent_read_limit,
+            "concurrentWriteLimit" if camel_case else "concurrent_write_limit": self.concurrent_write_limit,
+            "concurrentDeleteLimit" if camel_case else "concurrent_delete_limit": self.concurrent_delete_limit,
+        }
```
frxstrem, line 125: Q: Why does this type implement `dump` itself instead of using the inherited implementations from `CogniteResource`?

cognite/neat/core/_client/_api/statistics.py
```diff
@@ -0,0 +1,91 @@
...
+            ProjectStatsAndLimits: The requested statistics and limits
+
+        Examples:
+            Fetch project statistics (and limits) and check the current number of data models vs.
+            and how many more can be created:
+                >>> from cognite.client import CogniteClient
+                >>> client = CogniteClient()
+                >>> stats = client.data_modeling.statistics.project()
+                >>> num_dm = stats.data_models.current
+                >>> num_dm_left = stats.data_models.limit - num_dm
```
frxstrem, line 41: Are these examples accurate? The relation to the written code isn't clear to me.

cognite/neat/core/_client/_api/statistics.py
frxstrem, line 1: > The reason this functionality is not included in the cognite-sdk is that the endpoint is not yet ready and should not be exposed to end users through the SDK. However, in this pull request, it is not directly exposed to the end user.

Could you elaborate a bit on this, as I'm a bit confused. What does "the endpoint is not ready" mean? What are the risks of exposing it? And if it is not exposed to end users, where is it exposed in this PR?

doctrino: The endpoint is currently hidden, not in alpha yet. Starbase does not want to do a change in the PySDK. 

This is not exposed in `neat` as it is internal. I will make another PR were I use it to verify before uploading instances, but do it carefully in-case there is a change to the endpoint in the future.

cognite/neat/core/_client/data_classes/statistics.py
doctrino: The default `dump` method you get from `CogniteResource` does not handle nested object. We do the same in the PySDK.

cognite/neat/core/_client/_api/statistics.py
doctrino: Yes, the `ProjectStatsAndLimits` object there is a `data_models` attribute which has `.current` and `.limit`.

frxstrem: I mean, it's not clear how it relates to the `StatisticsAPI` class that it's an example for? Presumably it should somehow be using that?

rsjr (dismissed):

frxstrem (commented):
:unicorn:

rsjr (approved):

PR #1168, opened 2025-05-07
Refactor YAMLImporter to Dict Importer

rsjr:
# Description

Refactoring YAMLImporter into DictImporter.This refactor enables to have other "dict like" importers such as plain JSON files, and also Python dict (not reading from a static file). 

Also, It introduces a base clase for file read in the importers, which can also be reused later on for spreadsheets or other sources.

I am skipping a version bump for now as there is no change to the end users, but the upcoming PRs will introduce new functionalites which will be possible because of this refactor

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip

### Improved
- Refactoring YAMLImporter into a generic DictImporter
- Having a BaseFileReader class so we can reuse this later


doctrino (approved):
Good work

nikokaoja (approved):

doctrino (approved):

cognite/neat/core/_rules/importers/_yaml2rules.py
```diff
@@ -6,16 +6,38 @@
...
+        filepath: Path, allowed_extensions: frozenset[str] = frozenset([".yaml", ".yml"])
+    ) -> tuple[dict[str, Any], list[NeatIssue]]:
+        """Read a YAML file and return the data with any issues encountered."""
+        issues = BaseFileReader.validate_file(filepath, allowed_extensions)
+        if issues:
+            return {}, issues
+        # Try to load the YAML
+        try:
+            data = yaml.safe_load(filepath.read_text())
+            if not isinstance(data, dict):
```
thorkildcognite: An empty file will give you None back from safe_load, so you could special case that to specify that it is because of an empty file instead of "is not a dictionary").

cognite/neat/core/_data_model/importers/_yaml2rules.py
```diff
@@ -0,0 +1,123 @@
...
+    ) -> tuple[dict[str, Any], list[NeatIssue]]:
+        """Read a YAML file and return the data with any issues encountered."""
+        issues = BaseFileReader.validate_file(filepath, allowed_extensions)
+        if issues:
+            return {}, issues
+        # Try to load the YAML
+        try:
+            data = yaml.safe_load(filepath.read_text())
+            if not isinstance(data, dict):
+                issues.append(FileReadError(filepath, "YAML content is not a dictionary"))
```
thorkildcognite, line 34: Could consider special casing empty file here for clarity.

thorkildcognite:
risk review ok

PR #1195, opened 2025-05-12
Neat 969 rename dms to physical data model

nikokaoja:
# Description

This is a terminology sync PR, focused on renaming `DMS Rules` to `Physical Data Model`.

**NOTE**: Large number of lines changes due to renaming, no new functionality. It is not possible to split this into smaller PRs.

## Bump

- [ ] Patch
- [ ] Minor
- [X] Skip

### Improved

- Renaming `DMS Rules` to `Physical Data Model`


doctrino (dismissed):

doctrino (approved):

doctrino (approved):

doctrino (approved):

rsjr (commented):
Padawan risk reviewer: LGTM. No changes at all, just renaming.

cognite/neat/core/_data_model/models/entities/_single_value.py
```diff
@@ -344,17 +347,20 @@ def load(cls: "type[T_DMSEntity]", data: Any, strict: Literal[True], **defaults:
     @classmethod
```
thorkildcognite, line 347: I gather having this as a string and turning of type checking here is because T_DMSEntity is defined later on, and moving it won't help either, since it is bound to PhysicalEntity? So you need that type to define the class that gives you the type?

It should probably be renamed to T_PhysicalEntity, though?

I am not tremendously happy about this, but this is not something to fix in this PR. I would like to have a ticket on fixing this (if possible) that we can track, though.

thorkildcognite:
@nikokaoja Do you need to update `tests/tests_integration/test_session/test_graph_flow/test_uplift_workflow_to_python.yml` too? It still has the old references to logical model, and not conceptual. 

But looking at it more closely, I'm not sure this data is part of automated tests, but instead something you use manually? Example is that the GraphData.am-raw_graph_ttl references `aml_raw_graph.ttl`, but that file name is now `aml-raw-graph.ttl` (notice the dashes).

cognite/neat/core/_data_model/models/entities/_single_value.py
nikokaoja: Great catch @thorkildcognite  ! I've renamed it to `T_PhysicalEntity` .

I must say we have a bit of over-use of `TypeVar` accross the code base, large due to screaming of mypy.
Nevertheless, we can create a ticket to go through and see how we can remove, hopefully, most of them.

nikokaoja:
> @nikokaoja Do you need to update `tests/tests_integration/test_session/test_graph_flow/test_uplift_workflow_to_python.yml` too? It still has the old references to logical model, and not conceptual.

This is regression data of an old test that had an unpredictable behavior that we temporarily suspended. Nevertheless, I have updated the regression data. 

> But looking at it more closely, I'm not sure this data is part of automated tests, but instead something you use manually? Example is that the GraphData.am-raw_graph_ttl references `aml_raw_graph.ttl`, but that file name is now `aml-raw-graph.ttl` (notice the dashes).

Correctly spotted, however, not all the tests got updated to use the new structure of test data folder that @doctrino introduced. If we did, the issue you spotted with the wrong file name would be captured earlier. Nevertheless, I am bulldozing through the entire code base and cleaning up lose ends. I will add this as a specific task to complete.

cognite/neat/core/_data_model/models/entities/_single_value.py
doctrino: You can replace it with `Self` now, that should make mypy happy.

thorkildcognite:
risk review ok

cognite/neat/core/_data_model/models/entities/_single_value.py
thorkildcognite: @nikokaoja I think this would be a good improvement and get rid of an ugly string!

PR #1204, opened 2025-05-15
Plugin system

nikokaoja:
# Description

Enables creation of plugins for neat

Example of external neat plugin [IMF Importer](https://github.com/thisisneat-io/imf-importer-plugin), which usage is demonstrated in [this notebooke](https://github.com/thisisneat-io/imf-importer-plugin/blob/main/examples/how-to-use-plugin.ipynb).

## Bump

- [ ] Patch
- [x] Minor
- [ ] Skip

### Added

- Plugin manager system architecture and implementation
- Private access point for plugins in NeatSession with clear message if plugin being used is external


cognite/neat/core/_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self, name: str, kind: type[T_Plugin], module_path: str, class_name: str):
+        self.name = name
+        self.kind = kind
+        self.module_path = module_path
+        self.class_name = class_name
+        self._class: type[T_Plugin] | None = None
+
+    def get_class(self) -> type[T_Plugin]:
+        if self._class is None:
+            module = __import__(self.module_path, globals(), locals(), [""])
```
doctrino, line 72: Suggest having som error handling here in case this goes wrong.

cognite/neat/core/_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+    "register",
+]
+
+# Here we configure entry points for external plugins to be registered
+# in the plugin manager.
+plugins_entry_points = {
+    "cognite.neat.core.plugins.data_model.importers": DataModelImporter,
+}
+
+_plugins: dict[tuple[str, type[Any]], InternalPlugin] = {}
```
doctrino, line 33: What is supposed to be in the `type[Any]`?

PR #1207, opened 2025-05-16
🐛Workaround Coupled Views

doctrino:
# Description

This is a bug in the DMS API which is already reported. This PR implements a temporary workaround.
 
## Risk Reviewer
Note the same workaround has already been through risk-review for Toolkit: https://github.com/cognitedata/toolkit/pull/1490

This is a 191 line PR, the rest is YAML - test data.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- The `neat.to.cdf.data_model()` no longer fails for larger models coupled views through reverse direct relations and implements.


nikokaoja (dismissed):

cognite/neat/core/_utils/tarjan.py
```diff
@@ -0,0 +1,44 @@
+from typing import TypeVar
+
+T = TypeVar("T")
+
+
+def tarjan(dependencies_by_id: dict[T, set[T]]) -> list[set[T]]:
```
larshagencognite, line 6: Tests are not sufficient to say this algorithm is correctly implemented. Have you considered taking a dependency to handle this rather than rolling your own?

larshagencognite: Another option is to invest in more testing

doctrino: I think adding another dependency for a 20 line algorithm that is here because of a bug in the API is not worth it.

Can you be more concrete on the testing? It is not obvious to me what is missing in the testsing.

https://github.com/cognitedata/neat/pull/1207/files/0c828d9ab40ea614186d52a7c7b389899373942e#diff-a12af745afb46d81cddb10bbdedbdea69a2e850547c25818580c4841d5b30f1bR7

larshagencognite: I think a property test is reasonable here.
For any random graph:
  For each pair of nodes: a,b
    a,b is in same connected component <=> they are mutually reachable

doctrino: Thanks for the clarification, that is helpful.

doctrino: I added a few more concrete tests as well as the property test you suggested above. Note I added the dependency `hypothesis` to assist with the property based testing.

larshagencognite (approved):
🦄

PR #1210, opened 2025-05-18
Fix naming Standardization (#1209)

doctrino:
# Description

Numbers gets removed in the to_camel function.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- The `neat.infer()` method no longer loses numbers in inferred properties when standardizing property IDs.


nikokaoja (approved):
Nice catch !

Just check failed test

PR #1218, opened 2025-05-20
[NEAT-1004] 🧑‍🏫 Support index cursorable

doctrino:
# Description

Validation for this new syntax is introduced in #1230 

**Context** The neat spreadsheet format is (with the exception of filters) a one-to-one with the DMS API for containers, views, and data models. We are currently missing setting `cursorable` on btree indexes. This PR introduces it.
image

Adds support for setting index. 
* Introduce a `ContainerIndexEntity` to store the information. This is following the same pattern that we use for example Units.
* Replace the `list[str]` with `list[ContainerIndexEntity]` for physical properties.
## Bump

- [ ] Patch
- [x] Minor
- [ ] Skip

### Added

- Support for setting `cursorable` when specifying an index on container properties.

### Fixed

- Neat can now specify and maintains the order of multi-property order on indices. You do this with the syntax, `btree:multipPropertyIntdex(order=1)` in the Index column of the physical data model spreadsheet.

### Changed

- When setting an index on a property(ies) with a listable container, Neat uses an `InvertedIndex` instead of a `BTreeIndex`.


cognite/neat/core/_data_model/models/entities/_single_value.py
```diff
@@ -549,3 +554,11 @@ def as_node_entity(self) -> DMSNodeEntity:
...
     def as_concept_entity(self) -> ConceptEntity:
         return ConceptEntity(prefix=self.prefix, suffix=self.suffix, version=self.version)
+
+
+class ContainerIndexEntity(ConceptualEntity):
+    type_: ClassVar[EntityTypes] = EntityTypes.container_index
+    prefix: _UndefinedType = Undefined
+    suffix: str
+    cursorable: bool = False
+    by_space: bool = Field(False, alias="bySpace")
```
doctrino, line 564: `bySpace` is currently not supported in the PySDK, which is why we cannot support it yet, but added it here as a placeholder ready once it is in the PySDK.

cognite/neat/core/_data_model/models/physical/_exporter.py
```diff
@@ -380,14 +381,19 @@ def _create_containers(
...
 
-            index_properties: dict[str, set[str]] = defaultdict(set)
+            index_properties: dict[tuple[ContainerIndexEntity, bool], set[str]] = defaultdict(set)
             for prop in container_properties:
                 if prop.container_property is not None:
                     for index in prop.index or []:
-                        index_properties[index].add(prop.container_property)
-            for index_name, properties in index_properties.items():
+                        index_properties[(index, prop.is_list or False)].add(prop.container_property)
+            for (index_entity, is_list), properties in index_properties.items():
```
nikokaoja, line 389: if the properties is listable, it can be only resolved as InvertedIndex, but what if user sets `cursorable=True` ?
Should we not fail ? As it is now, we fix this silently without informing users that if they set cursorable to `True` this will have no effect. 

Same would be when tomorrow cognite-sdk starts supporting `bySpace`

I think we need to fail this in validation, so user is aware that users set something that is not possible.

cognite/neat/core/_data_model/models/physical/_unverified.py
```diff
@@ -135,7 +136,7 @@ class UnverifiedPhysicalProperty(UnverifiedComponent[PhysicalProperty]):
     default: str | int | float | bool | dict | None = None
     container: str | None = None
     container_property: str | None = None
-    index: str | list[str] | None = None
+    index: str | list[str | ContainerIndexEntity] | ContainerIndexEntity | None = None
```
nikokaoja, line 139: Should this be explicit type of filters instead of generic `ContainerIndexEntity` ? 

`BtreeIndexEntity` | `InvertedIndexEntity`

docs/excel_data_modeling/physical/properties.md
```diff
@@ -38,7 +38,28 @@ There are two `Value Type` that supports extra parameters
...
 
-### Connection Property
+### Index
+
+For data properties, you can specify an index. The index is used to speed up queries on the property. The syntax is as
+follows;
+
+| View        | View Property | Value Type | Container      | Container Property | Index                      |
+|-------------|---------------|------------|----------------|--------------------|----------------------------|
+| WindTurbine | name          | text       | GeneratingUnit | name               | nameIndex(cursorable=True) |
```
nikokaoja, line 48: Worth mentioning the short form of expressing Indexes:

`external_id_of_filter(properties...)`

nikokaoja (changes requested):
Two main comments:

- Need for explicit filter entities for the two filter types
- We should fail in validation if someone is setting BTreeImdex filter while having listable properties.

cognite/neat/core/_data_model/models/physical/_validation.py
```diff
@@ -391,7 +391,9 @@ def _consistent_container_properties(self) -> IssueList:
                         "rows",
                     )
                 )
-            index_definitions = {",".join(prop.index) for _, prop in properties if prop.index is not None}
+            index_definitions = {
```
nikokaoja, line 394: Are we failing validation if someone use same index external id (known as index-identifier in DMS) for different properties of the same container?

E.g.

Container Property Index
MyCont    name          nameIndex
MyCont    altName     nameIndex(cursorable=True)   


I guess ids of filters must be unique within a container

cognite/neat/core/_data_model/models/entities/_single_value.py
```diff
@@ -549,3 +554,11 @@ def as_node_entity(self) -> DMSNodeEntity:
 
     def as_concept_entity(self) -> ConceptEntity:
         return ConceptEntity(prefix=self.prefix, suffix=self.suffix, version=self.version)
+
+
```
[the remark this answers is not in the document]
doctrino: I solved it with a Literal.

cognite/neat/core/_data_model/models/physical/_validation.py
nikokaoja: > Yes, it can fail for this reason, suggest we add validation in a separate PR. Agree

nikokaoja (dismissed):

tests/tests_unit/test_rules/test_models/test_entities.py
```diff
@@ -24,6 +25,36 @@
...
+    (
+        ContainerIndexEntity,
+        "name",
+        ContainerIndexEntity(suffix="name"),
+    ),
+    (
+        ContainerIndexEntity,
+        "name(cursorable=True)",
+        ContainerIndexEntity(suffix="name", cursorable=False),
+    ),
```
thorkildcognite, line 37: I don't think this test works. I ran it locally, and no matter what I changed cursorable to, the test did not fail. Only if I changed name, it would fail.

tests/tests_unit/test_rules/test_models/test_entities.py
```diff
@@ -24,6 +25,36 @@
...
+    (
+        ContainerIndexEntity,
+        "name(cursorable=True)",
+        ContainerIndexEntity(suffix="name", cursorable=True),
+    ),
+    (
+        ContainerIndexEntity,
+        "name(bySpace=True,cursorable=True)",
+        ContainerIndexEntity(suffix="name", cursorable=True, bySpace=True),
```
thorkildcognite, line 46: Same as above. If I changed byspace and fiddled, it did not fail.

tests/tests_unit/test_rules/test_models/test_entities.py
thorkildcognite: To trigger a failure, you have to set strict to True in the call to _load in line 164.

I suggest you also add a non-happy-path test to verify this.

thorkildcognite (changes requested):
Please review my comments, and I think there is something off with the tests in test_entities.py.

docs/excel_data_modeling/physical/properties.md
```diff
@@ -38,7 +38,37 @@ There are two `Value Type` that supports extra parameters
...
+| View        | View Property | Value Type | Container      | Container Property | Index                            |
+|-------------|---------------|------------|----------------|--------------------|----------------------------------|
+| WindTurbine | name          | text       | GeneratingUnit | name               | btree:nameIndex(cursorable=True) |
+
+Note that you can specify whether the index is cursorable or not, see 
+[here](https://docs.cognite.com/cdf/dm/dm_guides/dm_performance_considerations/#pagination-cursorable-indexes).
+
+You can also specify indices that uses multiple data properties. You can do this by specifying the same `Index` name
+for multiple properties. See the example below:
+
```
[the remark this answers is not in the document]
doctrino: This is a good catch. Introducing new syntax to handle it.

tests/tests_unit/test_rules/test_models/test_entities.py
doctrino: Added test, found the bug, fixed it, and updated the test data.

cognite/neat/core/_data_model/models/entities/_single_value.py
```diff
@@ -549,3 +549,16 @@ def as_node_entity(self) -> DMSNodeEntity:
...
+class ContainerIndexEntity(ConceptualEntity):
+    type_: ClassVar[EntityTypes] = EntityTypes.container_index
+    prefix: _UndefinedType | Literal["btree", "inverted"] = Undefined
```
nikokaoja, line 556: What kind of index is Undefined? it can only be `betree` or `inverted` , we should not allow Undefined. I am aware that this was an easiest way to keep backward compatibility with old data models (where type of index was not explicitly stated, but implicitly configured). However, we know exactly what type of indexes there are, and we should not allow for "undefined" type. On importing, we can convert from "non-typed" to "typed" index and raise warning to user that we did this for them, and that in future they are expected to strongly type indexes.

cognite/neat/core/_data_model/models/physical/_validation.py
```diff
@@ -391,7 +391,9 @@ def _consistent_container_properties(self) -> IssueList:
...
+                ",".join([index.suffix for index in prop.index]) for _, prop in properties if prop.index is not None
```
nikokaoja, line 395: if type of index is not explicitly set, we need to raise warning or we need to clearly state in docs that if one is not setting this, we will automatically set it to the default index type.

nikokaoja (changes requested):

cognite/neat/core/_data_model/models/entities/_single_value.py
doctrino: Discusse agree that we will give a warning. This means that the code here will be kept as is, and in #1230 will introduce the warning.

tests/tests_unit/test_rules/test_models/test_entities.py
```diff
@@ -24,6 +25,36 @@
```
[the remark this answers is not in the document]
doctrino: After discussion, we allow this not to be set, thus `ContainerIndexEntity` should no fail if it is not provided but function as today were `Undefined` means the user has not set an index type. Then, in the validation PR #1230, I will add a warning if the index type is undefined.

nikokaoja (approved):

thorkildcognite:
risk review ok

PR #1219, opened 2025-05-20
Update pyproject.toml with right optional dependency group

nikokaoja:
# Description

Currently optional dep. group is called `pyoxigraph`, however in our code base we refer to it as `oxi`. 

## Bump

- [X] Patch
- [ ] Minor
- [ ] Skip

### Changed

- Reverted the group name from `pyoxigraph` to `oxi`.


rsjr (approved):
risk review ok

PR #1233, opened 2025-06-06
Neat 1031 plugin access via NeatSession

nikokaoja:
# Description

This PR provides interface in NeatSession for accessing external plugins (if they exist). If external plugin does not exist a nice error is returned:
Screenshot 2025-06-06 at 15 12 51

Example of WIP external plugins:
- [IMF Importer](https://github.com/thisisneat-io/imf-importer-plugin)
- [CFIHOS](https://github.com/cognitedata/industrial-standard-processor)

To reviewers: this is stacked PR on top of [Neat-1030](https://github.com/cognitedata/neat/pull/1232).

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- Access to the neat plugins via NeatSession


doctrino (approved):

cognite/neat/session/_plugin.py
```diff
@@ -0,0 +1,65 @@
...
+        !!! note "kwargs"
+            Users must consult the documentation of the plugin to understand
+            what keyword arguments are supported.
+        """
+        warnings.filterwarnings("default")
+        ExperimentalFlags.plugin.warn()
+
+        # Some plugins may not support the io argument
+        if io:
+            reader = NeatReader.create(io)
```
einarmo, line 48: While the typing of `NeatReader` is indeed `Any`. The type is in practice restricted to `str | Path`. Having what appears to be a fairly critical API use `Any` as type annotation isn't great. Since this is now becoming part of a public API, it's even more important to type it correctly.

cognite/neat/session/_plugin.py
```diff
@@ -0,0 +1,65 @@
...
+            path = reader.materialize_path()
+        else:
+            path = None
+
+        self._state._raise_exception_if_condition_not_met(
+            "Data Model Read",
+            empty_data_model_store_required=True,
+        )
+
+        plugin = plugin_manager.get(name, DataModelImporterPlugin)
```
einarmo, line 58: Typing breaks down here, since this method returns `Any`. Can you not let `get` be generic and return the requested type?

cognite/neat/session/_plugin.py
```diff
@@ -0,0 +1,65 @@
...
+
+        print(
+            f"You are using an external plugin {plugin.__name__}, which is not developed by the NEAT team."
+            "\nUse it at your own risk."
+        )
+
+        return self._state.data_model_import(plugin().configure(io=path, **kwargs))
```
einarmo, line 65: `configure` also has the same issue, effectively. At this point `io` is `str | Path`, and `configure should probably reflect this.

I'd recommend you define a `NeatIO` type alias, or something like that, since this type appears to be used quite widely.

tests/tests_unit/test_session/test_plugin.py
```diff
@@ -0,0 +1,19 @@
...
+
+    # Neat Session does not raise an error, but prints it.
+    output = io.StringIO()
+    with contextlib.redirect_stdout(output):
+        neat.plugins.data_model.read("csv", "./test.txt")
+
+    printed_statements = output.getvalue()
+    assert printed_statements == (
+        "[ERROR] PluginError: No plugin of type 'DataModelImporterPlugin' registered \nunder name 'csv'\n"
+    )
```
einarmo, line 19: I'd like to see a happy-path test here as well, i.e. actually run the `read` method to the end. Unless this is tested elsewhere?

nikokaoja: @einarmo those tests will be added when external plugins are published on PyPI. Otherwise, this requires to build a mock external plugin python package in side of neat test to demo.

cognite/neat/session/_plugin.py
nikokaoja: @einarmo 100% agree with you! I will add dedicated Jira ticket to clean up this and force proper typing.

For time being I will this in plugin endpoint , i.e. setting `io` to ` str | Path | None` as I anticipate some plugins not to take any `io` what so ever (e.g. CFIHOS importer plugin will already have).

tests/tests_unit/test_session/test_plugin.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_plugin_read(mock_external_plugin):
```
nikokaoja, line 44: @einarmo added requested test

einarmo:
:unicorn:

doctrino (approved):

PR #1242, opened 2025-06-20
[NEAT-1115]  Unbound local error

doctrino:
# Description

See below

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Running `neat.to.cdf.data_model()` no longer gives a `UnboundLocalError` if there is an error in the deployment.


rsjr (commented):
risk review ok

PR #1243, opened 2025-06-22
[NEAT-1122] Inferred indices

doctrino:
# Description

In the conversion of index from phycial model to DMS schema, I assumed in the original implementation that `Undefined` is a singleton and it turns out that is not the case.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Neat no longer silently fails for `neat.to.cdf.data_model(...)` when indices does not specify `btree` or `inverted`. Note this bug was introduced in `0.123.0`


cognite/neat/core/_data_model/models/physical/_exporter.py
```diff
@@ -391,9 +391,9 @@ def _create_containers(
                 container.indexes = container.indexes or {}
                 index_property_list = [prop_id for prop_id, _ in sorted(properties, key=lambda x: x[1].order or 0)]
                 index_entity = properties[0][1]
-                if index_entity.prefix == "inverted" or (index_entity.prefix is Undefined and is_list):
+                if index_entity.prefix == "inverted" or (index_entity.prefix == Undefined and is_list):
                     container.indexes[index_name] = InvertedIndex(properties=index_property_list)
-                elif index_entity.prefix == "btree" or (index_entity.prefix is Undefined and not is_list):
+                elif index_entity.prefix == "btree" or (index_entity.prefix == Undefined and not is_list):
```
doctrino, line 396: Fix

rsjr (approved):
good stuff

PR #1244, opened 2025-07-01
Relax regex on concept id

nikokaoja:
# Description

This is first of multiple stacked PRs that are dealing with relaxation of conceptual data model, which aim at lossless importing and inferring of conceptual data model.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Changed

- Reduce concept id regex restriction allowing additional characters


rsjr (commented):
risk review ok

PR #1246, opened 2025-07-14
Improve OWL importer by quoting IDs

nikokaoja:
# Description

OWL importer has been improved allowing for class and property ids which contain "non-web-safe" characters, which can exist since OWL allows defining ids using IRIs , which are allow for more characters than URIs.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- OWL importer got improved flexibility for ids of entities that making ontology

### DIFF
--- a/cognite/neat/core/_data_model/importers/_rdf/_shared.py
+++ b/cognite/neat/core/_data_model/importers/_rdf/_shared.py
@@ -1,4 +1,5 @@
 from typing import cast
+from urllib.parse import quote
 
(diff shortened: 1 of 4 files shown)


doctrino (approved):
I think this is good. But could this be something we do in the unverified -> verified conceptual model? Then we do not have to remember it for every importer.

doctrino (approved):

rsjr (commented):
risk review ok

PR #1247, opened 2025-07-15
Enable concept only conceptual data models

nikokaoja:
# Description

This enables conceptual data models to consist of only concepts. It raises nice warning.

Risk reviewers: This is stacked PR on top of [1246](https://github.com/cognitedata/neat/pull/1246)

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- Concept only conceptual data models are now allowed

PR #1248, opened 2025-07-15
Enable defining properties in conceptual data model which are not defined for any concept (dangling properties)

nikokaoja:
# Description

This PR adds support for defining dangling properties—properties in the conceptual data model that are not yet associated with any specific concept.

**Justification**:
When generating a conceptual data model from an OWL ontology, it’s common to encounter properties (e.g., owl:ObjectProperty, owl:DatatypeProperty) that are declared but not explicitly linked to any defined class via domain or range. Supporting dangling properties ensures these are not lost during transformation and allows the model to faithfully represent the source ontology structure. It also enables incremental refinement, where associations can be added later as understanding of the domain evolves.

**Note**: This PR is stacked on top of #1247

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- Added support for dangling properties in conceptual models, useful for OWL imports with undefined domains/ranges.


rsjr (commented):
risk review ok

PR #1249, opened 2025-07-16
Update excel importer to handle dangling properties

nikokaoja:
# Description

Updated the Excel importer to handle dangling properties — properties in the Excel sheet that are not associated with any Concept. These are now captured and processed gracefully rather than causing import errors.

Risk Reviewers: This is stacked PR

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- Improved Excel importer to handle dangling properties gracefully.


doctrino (dismissed):

cognite/neat/core/_utils/spreadsheet.py
```diff
@@ -98,6 +98,10 @@ def read_individual_sheet(
         # Special handling for Value Type column, #N/A is treated specially by NEAT it means Unknown
         raw["Value Type"] = raw["Value Type"].replace(float("nan"), "#N/A")
 
+    if "Concept" in raw.columns:
```
rsjr, line 101: "Concept" is always hardcoded like this or is it possible to have variations (small caps, all caps etc etc)?

cognite/neat/core/_utils/spreadsheet.py
```diff
@@ -98,6 +98,10 @@ def read_individual_sheet(
...
+        # Special handling for Concept column, #N/A is treated specially by NEAT it means Unknown
+        raw["Concept"] = raw["Concept"].replace(float("nan"), "#N/A")
```
rsjr, line 103: see above

rsjr (commented):
1 minor comment

cognite/neat/core/_utils/spreadsheet.py
nikokaoja: Yes it is, as otherwise it would not be NEAT Excel template

PR #1252, opened 2025-07-18
Remove unnecessary method and check

nikokaoja:
# Description

Previously we allowed for re-definition of properties. This has not been case for a long time. However, a method and check that looking for the redefinitions pertained when a conceptual data model is exported to ontology. This PR removes them as they are no longer necessary.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Removed

- Remove method and check for properties redefinitions for ontology exporter

### DIFF
--- a/cognite/neat/core/_data_model/exporters/_data_model2ontology.py
+++ b/cognite/neat/core/_data_model/exporters/_data_model2ontology.py
@@ -17,10 +17,8 @@ from cognite.neat.core._data_model.models.conceptual import (
     ConceptualMetadata,
     ConceptualProperty,
 )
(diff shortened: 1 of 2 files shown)


doctrino (approved):
I would change the wording in the changelog. Removing a method is an implementation detail, it is not what changes for the user. 

> Remove method and check for properties redefinitions for ontology exporter
If I understood this correctly, the change to the user is:
Allow for redefinition of properties in ontology exporter.

psalaberria002 (approved):
:unicorn:

doctrino (approved):

PR #1257, opened 2025-07-25
GraphToDataModel importer

doctrino:
# Description

**Context**: Neat has two main flows.
1. Going from some data model source -> into Neat data model format -> DMS
2. Going from some data source -> into Neat Triple store -> DMS.

This Importer connects the first two, it allows to go from the internal Neat triple store into the Neat data model format, essentially inferring a data model from a data. This PR does not expose this to the end-user of neat, so there is no bump in this PR.

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite/neat/core/_data_model/importers/_graph2data_model.py
```diff
@@ -0,0 +1,288 @@
...
+            for result_row in self.store.dataset.query(property_query):
+                property_uri, value_type_uri = cast(tuple[URIRef, URIRef], result_row)
+                if property_uri == RDF.type:
+                    continue
+                occurrence_query = self._MAX_OCCURRENCE_QUERY.format(type=type_uri, property=property_uri)
+                max_occurrence = 1  # default value
+                occurrence_results = list(self.store.dataset.query(occurrence_query))
+                if occurrence_results and occurrence_results[0] and occurrence_results[0][0]:
+                    max_occurrence_literal = cast(RdfLiteral, occurrence_results[0][0])
+                    max_occurrence = int(max_occurrence_literal.toPython())
```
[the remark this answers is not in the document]
doctrino: Will defer this to a different task as that will significantly increase the complexity of this PR.

cognite/neat/core/_data_model/importers/_graph2data_model.py
```diff
@@ -0,0 +1,288 @@
...
+                properties_by_concept_by_property[concept_uri][read_prop.property_uri].append(read_prop)
+        return properties_by_concept_by_property
+
+    @staticmethod
+    def _create_concept(
+        type_uri: URIRef, set_instance_source: bool, prefixes: dict[str, Namespace], implements: str | None = None
+    ) -> tuple[str, UnverifiedConcept]:
+        namespace, suffix = split_uri(type_uri)
+        if namespace not in prefixes:
+            prefixes[namespace] = Namespace(namespace)
```
nikokaoja, line 229: Considering we get following type_uris:
- http://anders.com/type_x
- http://nikola.com/type_y

our resulting Prefixes "table" will eventually look like this:

| Prefix   | Namespace |
| -------- | ------- |
| http://anders.com/  | http://anders.com/   |
| http://nikola.com/ | http://nikola.com/  |

which will be illformed since we want that table in form of:

| Prefix   | Namespace |
| -------- | ------- |
| prefix_1  | http://anders.com/   |
| prefix_2 | http://nikola.com/  |


we have method [_add_uri_namespace_to_prefixes](https://github.com/cognitedata/neat/blob/1b7afb3d7e0b1bfc07cca7f09462e0e824a558df/cognite/neat/core/_data_model/importers/_rdf/_base.py#L139) that can help for this.

cognite/neat/core/_data_model/importers/_graph2data_model.py
```diff
@@ -0,0 +1,288 @@
...
+        count_by_type = self._read_types_with_counts_from_graph()
+        if not count_by_type:
+            warnings.warn(
+                NeatValueWarning("Cannot infer data model. No RDF.type triples found in the graph."), stacklevel=2
+            )
+            return ImportedDataModel(UnverifiedConceptualDataModel(metadata, [], [], {}), {})
+
+        read_properties = self._read_concept_properties_from_graph(count_by_type, parent_by_child)
+
+        prefixes: dict[str, Namespace] = {}
```
nikokaoja, line 110: Suggestion, start of with default set of prefixes :

prefixes = get_default_prefixes_and_namespaces()

see [old one](https://github.com/cognitedata/neat/blob/1b7afb3d7e0b1bfc07cca7f09462e0e824a558df/cognite/neat/core/_data_model/importers/_rdf/_inference2rdata_model.py#L381)

tests/data/_graph/car.py
```diff
@@ -30,15 +30,15 @@
         (_instance_ns["Blue"], RDF.type, _model_ns["Color"]),
         (_instance_ns["Blue"], _model_ns["name"], Literal("blue")),
```
nikokaoja, line 31: suggestion, augment this test and make:

TYPE_SPACE = "sp_types"
PROPERTY_SPACE = "sp_properties"
_type_ns = Namespace(DEFAULT_SPACE_URI.format(space=TYPE_SPACE))
_property_ns = Namespace(DEFAULT_SPACE_URI.format(space=PROPERTY_SPACE))

and then update triples to following form:

```
(_instance_ns["Car1"], RDF.type, _type_ns["Car"]),
(_instance_ns["Car1"], _property_ns["Car.Manufacturer"], _instance_ns["Toyota"]),
...
```

resulting conceptual data model should then have one or more entires in Prefixes sheet.

nikokaoja (approved):
Really readable implementation! This was pleasure to read.

Though we are no longer using short form `prefix:suffix` to express `instance_source`, it is worth updating code to properly capture all the prefixes we pull from graph.

Alt. would be to drop Prefixes sheet from Conceptual Data Model all toghether

cognite/neat/core/_data_model/importers/_graph2data_model.py
doctrino: I did not include it intentionally. My thinking is that if we include it, you force the user of this importer to have the default prefixes and you do not know which ones originated from the graph. It is my attempt to try to learn from past mistakes to do too much in an importer. 

As a compromise, what about having a argument in the init method `include_default_prefixes=True`?

cognite/neat/core/_data_model/importers/_graph2data_model.py
doctrino: will do

tests/data/_graph/car.py
doctrino: will do

cognite/neat/core/_data_model/importers/_graph2data_model.py
nikokaoja: Agree with you, for time being we do not need to offer adding default prefixes, if we needed we can added it later.

nikokaoja (approved):

PR #1258, opened 2025-07-25
Generic context on Imported Data Model

doctrino:
# Description

`ImportedDataModel` is an internal data structure we use to import data models from different sources such as spreadsheet, CDF, ontologies and so on. We use the context today when the source is a spreadsheet to tell the user exactly which row have warnings/errors to make it easy for them to fix. This is similar to a `metadata` field used in the CDF API.

This PR changes the type of context from one that is only valid for a spreadsheet source to a more generic one. This is because in an upcoming PR we will use the context in a different importers that are not importing form spreadsheet, but for example a Graph or Ontology. Then we will use this context for different purposes.  

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite/neat/core/_data_model/_shared.py
+++ b/cognite/neat/core/_data_model/_shared.py
@@ -1,3 +1,4 @@
+from collections.abc import Mapping
 from dataclasses import dataclass
 from typing import Generic, TypeAlias, TypeVar
 
@@ -11,7 +12,6 @@ from cognite.neat.core._data_model.models.conceptual._unverified import (
 from cognite.neat.core._data_model.models.physical._unverified import (
     UnverifiedPhysicalDataModel,
 )
-from cognite.neat.core._utils.spreadsheet import SpreadsheetRead
 
 VerifiedDataModel: TypeAlias = ConceptualDataModel | PhysicalDataModel
 
@@ -32,7 +32,7 @@ class ImportedDataModel(Generic[T_UnverifiedDataModel]):
     """
 
     unverified_data_model: T_UnverifiedDataModel | None
-    context: dict[str, SpreadsheetRead]
+    context: Mapping[str, object]
 
     @classmethod
     def display_type_name(cls) -> str:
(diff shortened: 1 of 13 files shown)


cognite/neat/core/_data_model/_shared.py
```diff
@@ -32,7 +32,7 @@ class ImportedDataModel(Generic[T_UnverifiedDataModel]):
     """
 
     unverified_data_model: T_UnverifiedDataModel | None
-    context: dict[str, SpreadsheetRead]
+    context: Mapping[str, object]
```
doctrino, line 35: Making this a mapping instead of dict to signal that this is intended to be Immutable when you consume it.

cognite/neat/core/_data_model/models/physical/_validation.py
```diff
@@ -81,15 +81,17 @@ def __init__(
...
         self._data_model = data_model
         self._client = client
         self._metadata = data_model.metadata
         self._properties = data_model.properties
         self._containers = data_model.containers
         self._views = data_model.views
-        self._read_info_by_spreadsheet = read_info_by_spreadsheet or {}
+        self._read_info_by_spreadsheet = {
+            key: value for key, value in (context or {}).items() if isinstance(value, SpreadsheetRead)
+        }
```
doctrino, line 94: In the validators, we convert the context to what we specifically need here.

nikokaoja (approved):
Great! We can than leverage context beyond spreadsheets

erlendvollset:
Going from a typed to an untyped context object doesn't seem like a good evolution. Why does it need to be untyped?

doctrino:
> Going from a typed to an untyped context object doesn't seem like a good evolution. Why does it need to be untyped?

Was the description not clear enough?

doctrino:
> > > Going from a typed to an untyped context object doesn't seem like a good evolution. Why does it need to be untyped?
> > 
> > 
> > Was the description not clear enough?
> 
> This part?
> 
> > This PR changes the type of context from one that is only valid for a spreadsheet source to a more generic one. This is because in an upcoming PR we will use the context in a different importers.
> 
> This doesn't tell me why it has to be untyped, no.

Updated description.

erlendvollset:
I still don't understand why it has to be untyped. Being typed and being "generic" or "used for different purposes" are hardly mutually exclusive. 

Is the problem that the context will have a different shape depending on where it's being instantiated? You can then have a common base class for all context types which you use in these signatures.

PR #1260, opened 2025-08-05
Object oriented NeatSession: Data Model - READ APIs

nikokaoja:
# Description

Object oriented NeatSession endpoint for reading of data models. Now users can read data models into neat session using the following approach:

```
neat.data_model.read(...)
```

making this more convenient  way of interacting with neat, as well external data model reader plugins.

To risk-reviewers: this is part of refactoring of neat session endpoints which will be followed by a soft migration of users. Current endpoints in NeatSession are more "function based". Initially this approach worked very well, however, a need for "object oriented" design of endpoints emerged over time. This refactoring of endpoints are critical to provide an easy access to external plugins as well.

**Risk reviewing**: This PR does not bring any new functionalities. Methods already exist but are accessible via functional approach in interaction with NeatSession. They are copied to be "object oriented" accessible.

**Large number of lines is due to (regression) test data!**

## Bump

- [X] Patch
- [ ] Minor
- [ ] Skip

### Added

- Object oriented NeatSession endpoint for reading of data models.


cognite/neat/session/_session/_data_model.py
```diff
@@ -0,0 +1,154 @@
...
+    def __init__(self, state: SessionState) -> None:
+        self._state = state
+        self.read = ReadAPI(state)
+
+
+@session_class_wrapper
+class ReadAPI:
+    def __init__(self, state: SessionState) -> None:
+        self._state = state
+
```
nikokaoja, line 29: All the methods below already exist so they are `ctrl+c` `ctrl+v` here

sighol: Where did you copy paste these from? I found some similarities in `cognite/neat/session/_read.py`, but it does not looks like a pure copy-paste.

This is the excel function diff:
```patch
--- _read.py.txt        2025-08-07 15:42:02.949239100 +0200
+++ _data_model.py.txt  2025-08-07 15:42:22.945813896 +0200
@@ -1,26 +1,18 @@
-    def __call__(self, io: Any, enable_manual_edit: bool = False) -> IssueList:
+    def excel(self, io: str | Path, *, enable_manual_edit: bool = False) -> IssueList:
         """Reads a Neat Excel Data Model to the data model store.
         The data model spreadsheets may contain conceptual or physical data model definitions.

             Args:
                 io: file path to the Excel sheet
                 enable_manual_edit: If True, the user will be able to re-import data model
-                    which where edit outside NeatSession
-
-            !!! note "Manual Edit Warning"
-                This is an alpha feature and is subject to change without notice.
-                It is expected to have some limitations and may not work as expected in all cases.
+                    which where edit outside of NeatSession
         """
         reader = NeatReader.create(io)
         path = reader.materialize_path()

-        if enable_manual_edit:
-            warnings.filterwarnings("default")
-            ExperimentalFlags.manual_data_model_edit.warn()
-        else:
             self._state._raise_exception_if_condition_not_met(
                 "Read Excel Data Model",
-                empty_data_model_store_required=True,
+            empty_data_model_store_required=not enable_manual_edit,
             )

         return self._state.data_model_import(importers.ExcelImporter(path), enable_manual_edit)
```

sighol (commented):
I don't have experience with Neat, so forgive me if the risk review is a bit slow 😅

tests/tests_end_to_end/test_rules_flow/test_excel_importer_to_yaml_new_endpoint.yml
```diff
@@ -0,0 +1,665 @@
+containers:
```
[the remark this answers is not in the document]
sighol: Ahh, you are using `pytest_regression`, and by default it uses a path based on the test name. Alright.

tests/tests_end_to_end/test_rules_flow.py
```diff
@@ -28,6 +28,21 @@ def test_excel_importer_to_yaml(self, data_regression: DataRegressionFixture) ->
...
+        neat = NeatSession(verbose=False)
+
+        neat.data_model.read.excel(DOC_RULES / "information-architect-david.xlsx")
+
+        neat.convert()
+
+        exported_yaml_str = neat.to.yaml()
+
+        exported_rules = yaml.safe_load(exported_yaml_str)
+        data_regression.check(exported_rules)
```
sighol, line 41: nit: Could have used custom base names so that you don't need to create a new file. I am not sure it is _better_ though 🤷‍♂️ 

https://pytest-regressions.readthedocs.io/en/latest/overview.html#generated-filenames

```suggestion
        data_regression.check(exported_rules, basename="test_excel_importer_to_yaml")
```

PR #1266, opened 2025-08-12
Handling of OWL restriction for domain and range

nikokaoja:
# Description

Often a property is not only defined for a single concept, but re-used for multiple concepts. In case of ontological modeling this is achieved by applying owl restrictions on rdfs:domain and rdfs:range when defining an owl property. Now we are able to parse these restrictions, which leads to step closer to lossless importing of ontologies into neat expression of data model.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- Parsing of OWL restrictions for domain and range when those as expressed as blank nodes (i.e. using owl:unionOf , owl:intersectionOf)


rsjr (commented):
risk review ok

rsjr (approved):

PR #1268, opened 2025-08-22
Enabling restrictions in conceptual data model: adding restriction entities 

nikokaoja:
# Description

This is first of a few PRs that are tackling importing of data model restrictions often found in ontologies. PR brings definition of two basic forms of restrictions:
- Concept property constraint on values
- Concept property constraint on cardinality and value type

In addition to the definition of the two new entities, a short form of how those entities are represented as a string is provided. 

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip

### Added

- Added new neat entities: `ConceptPropertyValueConstraint` and `ConceptPropertyCardinalityConstraint` setting foundation for parsing of OWL restrictions

### DIFF
--- a/cognite/neat/core/_data_model/_constants.py
+++ b/cognite/neat/core/_data_model/_constants.py
@@ -43,6 +43,10 @@ class EntityTypes(StrEnum):
     prefix = "prefix"
     space = "space"
     container_index = "container_index"
+    concept_restriction = "conceptRestriction"
+    value_constraint = "valueConstraint"
+    cardinality_constraint = "cardinalityConstraint"
+    named_individual = "named_individual"
 
 
 def get_reserved_words(
--- a/cognite/neat/core/_data_model/models/entities/_restrictions.py
+++ b/cognite/neat/core/_data_model/models/entities/_restrictions.py
@@ -1,0 +1,210 @@
+import re
+from abc import ABC
+from typing import Any, ClassVar, Final, Literal, TypeAlias, TypeVar, cast, get_args
+
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+from rdflib import Literal as RDFLiteral
+from rdflib import URIRef
+
+from cognite.neat.core._data_model._constants import EntityTypes
+from cognite.neat.core._data_model.models.data_types import _XSD_TYPES, DataType
+from cognite.neat.core._data_model.models.entities._constants import _PARSE
+from cognite.neat.core._data_model.models.entities._single_value import ConceptEntity, ConceptualEntity
+from cognite.neat.core._issues.errors._general import NeatValueError
+from cognite.neat.core._utils.rdf_ import remove_namespace_from_uri
+
+ValueConstraints = Literal["allValuesFrom", "someValuesFrom", "hasValue"]
+CardinalityConstraints = Literal["minCardinality", "maxCardinality", "cardinality", "qualifiedCardinality"]
+
+
+VALUE_CONSTRAINT_REGEX = re.compile(
+    rf"^{EntityTypes.value_constraint}:(?P<property>[a-zA-Z0-9._~?@!$&'*+,;=%-]+)\((?P<constraint>{'|'.join(get_args(ValueConstraints))}),(?P<value>.+)\)$"
+)
+
+CARDINALITY_CONSTRAINT_REGEX = re.compile(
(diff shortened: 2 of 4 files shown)


tests/tests_unit/test_rules/test_models/test_restrictions.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+    def test_parse_invalid_restrictions(self, invalid_data):
+        with pytest.raises(NeatValueError, match="Invalid restriction format"):
+            parse_restriction(invalid_data)
+
+
+class TestNamedIndividualEntity:
+    def test_reset_prefix(self):
+        entity = NamedIndividualEntity(suffix="test", prefix="one-that-gets-reset")
+        entity.reset_prefix()
```
doctrino, line 142:
```suggestion
```

It is uncommon to call model_validator methods explicitly. They are invoked automatically. You can remove it without the test failing.

cognite/neat/core/_data_model/models/entities/_restrictions.py
```diff
@@ -0,0 +1,231 @@
...
+    # If neither pattern matches, provide a clear error
+    raise NeatValueError(
+        f"Invalid restriction format: {data}. "
+        f"Expected format: '{EntityTypes.value_constraint}:property(constraint,value)' "
+        f"or '{EntityTypes.cardinality_constraint}:property(constraint,value[,on])'"
+    )
+
+
+ConceptRestriction: TypeAlias = ConceptPropertyValueConstraint | ConceptPropertyCardinalityConstraint
+T_ConceptRestriction = TypeVar("T_ConceptRestriction", bound=ConceptRestriction)
```
doctrino, line 231: Both of these are subclassing `ConceptPropertyRestriction`. Why no just use that as a bound? Then, you subclass `ConceptPropertyRestriction` later then, they will automatically be included. 

```suggestion
T_ConceptRestriction = TypeVar("T_ConceptRestriction", bound=ConceptPropertyRestriction)
```

cognite/neat/core/_data_model/models/entities/_restrictions.py
```diff
@@ -0,0 +1,231 @@
...
+            return NotImplemented
+        return self.as_tuple() < other.as_tuple()
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, ConceptPropertyRestriction):
+            return NotImplemented
+        return self.as_tuple() == other.as_tuple()
+
+    def __hash__(self) -> int:
+        return hash(str(self))
```
doctrino, line 115: We do not have any tests that hits this. What about a set tests? Ensure that if you add the same constraint twice to a set it only have one?

tests/tests_unit/test_rules/test_models/test_restrictions.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+    def test_parse_basic_cardinality_constraints(
+        self, data, expected_property, expected_constraint, expected_value, expected_on
+    ):
+        result = ConceptPropertyCardinalityConstraint.load(data)
+
+        assert result.property_ == expected_property
+        assert result.constraint == expected_constraint
+        assert result.value == expected_value
+        assert result.on == expected_on
```
doctrino, line 89: Suggest to also assert that the serialization works in reverse as well. 
```suggestion
        assert result.on == expected_on
        
        assert str(result) == data
```

doctrino (commented):
A few suggestions.

doctrino (approved):

cognite/neat/core/_data_model/_constants.py
```diff
@@ -43,6 +43,10 @@ class EntityTypes(StrEnum):
     prefix = "prefix"
     space = "space"
     container_index = "container_index"
+    concept_restriction = "conceptRestriction"
+    value_constraint = "valueConstraint"
+    cardinality_constraint = "cardinalityConstraint"
```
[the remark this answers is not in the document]
nikokaoja: For simplicity we have currently one single EntityTypes vocabulary where we hold all the entities we are working with. I think we can still live with singular vocab.

cognite/neat/core/_data_model/models/entities/_restrictions.py
```diff
@@ -0,0 +1,230 @@
...
```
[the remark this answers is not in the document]
nikokaoja: We have same implementation for `ConceptEntity` and so far we have not bump into any issues.

PR #1272, opened 2025-09-15
Remove upper bound on packaging

Magssch:
# Description

This third-party dependency had an upper bound that caused compatibility issues with `cognite-toolkit` due to differing upper and lower bounds. The upper bound was recently removed with this change in `cognite-toolkit`: https://github.com/cognitedata/toolkit/commit/2d0ef1383325e2a074e607b9fbaf06fc24a0e1de

This commit removes the upper bound in Neat as well, ensuring both packages can now be installed in the same repo. Verified locally with an install with `cognite-toolkit=0.6.26` and using this branch as target for `cognite-neat`.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Removed upper bound on third-party dependency `packaging`. This makes
Neat compatible with toolkit.

- My change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "networkx>=3.4.2, <4.0.0",
     "mixpanel>=4.10.1, <5.0.0",
     # Used for loading neat engine
-    "packaging>=22.0, <25.0",
+    "packaging>=22.0",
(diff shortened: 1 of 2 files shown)


thorkildcognite:
risk review ok

PR #1283, opened 2025-09-21
[THIS-729] 🐈‍⬛ Add space pydantic

doctrino:
# Description

**Context**: We are changing the internal structure of the neat data models. We are dropping the physical data model format in Neat, and instead will have a one-to-one DMS format identical to the API. In principle, we could use the data classes that are in the `cognite-sdk`, but we decided to reimplement them in `pydantic`. This is to 1. Get better validation. 2. Better performance on serialization (pydantic does serialization in rust), 3. Make them easy to extend and connect to the Neat Conceptual model.

This PR implements Pydantic classes for Space. 

Notet this PR also creates the structure and testing for these new classes. That means the following changes as well

* Added dev dependency `polyfactory` to generate fake data in tests.
* Added auxiliary method `get_concrete_subclasses` that is used to dynamically find all the Resource implementations (space, containers, views, etc.) and test their `.as_request` method. **Reviewer**: Note this method comes from Toolkit and has been through risk-review before. 

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -148,3 +148,29 @@ jobs:
             fs.writeFileSync('last_git_message.txt', description);
       - name: Validate Changelog section in PR description
         run: uv run python dev.py changelog
+
+  build_and_test_jupyter_pyodide:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Install uv and set the python version
+        uses: astral-sh/setup-uv@v6
+        with:
+          python-version: ${{ env.MAIN_PYTHON_VERSION }}
+
+      - name: Build package using uv
+        run: uv build
+      - name: Set up Node.js
+        uses: actions/setup-node@v5
+        with:
+          node-version: "22"
+      - name: Install dependencies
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -126,12 +127,15 @@ allow_redefinition = true
...
+# We use reflection a lot in this codebace, and mypy does not understand that
+# finding all subclasses of an abstract class gives you concrete implementations.
+# Thus, we disable the type-abstract error.
+disable_error_code = ["type-abstract"]
 
 [tool.pydantic-mypy]
 init_forbid_extra = true
 init_typed = true
-warn_required_dynamic_aliases = true
+warn_required_dynamic_aliases = false
```
doctrino, line 138: In the new pydantic classes we want to use dynamic alias. This ensures that all alias are camel case, instead of manually setting it.

pyproject.toml
```diff
@@ -95,6 +95,7 @@ dev = [
     "tqdm>=4.67.1, <5",
     "deepdiff>=8.2.0, <9.0.0",
     "hypothesis>=6.135.0, <7.0.0",
+    "polyfactory>=2.22.2",
```
doctrino, line 98: This package helps with creating mock data in testing.

PR #1293, opened 2025-09-25
🚄Added accelerate team owner

doctrino:
# Description

I am adding the accelerate team and will use a separate PR to remove `iknow`. This is to verify it works before removing to avoid getting into a state were the repo has no owners in case the verification fails.

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip


rsjr (approved):
risk review ok
🚤

PR #1304, opened 2025-09-29
THIS 752: Correct target class and value type setting

nikokaoja:
# Description

This is stacked PR on top of #1303 

During the testing of SHACL engine which was taking shacl rules generated by neat it has beed detected that we were setting `NodeShape`'s `targetClass` and `PropertyShape`'s `node` to URI of SHACL shapes. The correct way is to set those properties to class URI, which in our case is stored as `instance_source` (if conceptual data model is linked to the instances) or to generate such URI from concept entity suffix and data model namespace. Furthermore, `PropertyShape`'s `path` was not leveraging `instance_source` of property (if it exists). The above has been implemented in this PR.

## Bump

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- Node shape target class and property shape node are now set to instance source (if it stated) or class uri

### DIFF
--- a/cognite/neat/v0/core/_data_model/exporters/_data_model2semantic_model.py
+++ b/cognite/neat/v0/core/_data_model/exporters/_data_model2semantic_model.py
@@ -618,7 +618,9 @@ class SHACLPropertyShape(_ModelConfig):
 
         return cls(
             id_=BNode(),
-            path=namespace[definition.property_],
+            path=definition.instance_source[0]
+            if definition.instance_source and len(definition.instance_source) == 1
+            else namespace[definition.property_],
             node_kind=SHACL.IRI if definition.type_ == EntityTypes.object_property else SHACL.Literal,
             expected_value_type=expected_value_type,
             min_count=definition.min_count,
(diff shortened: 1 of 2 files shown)


cognite/neat/v0/core/_data_model/exporters/_data_model2semantic_model.py
```diff
@@ -548,21 +550,28 @@ def triples(self) -> list[tuple]:
...
+
+        if concept.implements:
+            parent = [namespace[str(parent.suffix) + "Shape"] for parent in concept.implements]
         else:
             parent = None
         return cls(
-            id_=namespace[f"{concept_definition.concept.suffix!s}Shape"],
-            target_class=namespace[str(concept_definition.concept.suffix)],
+            id_=namespace[f"{concept.concept.suffix!s}Shape"],
+            target_class=concept.instance_source or namespace[str(concept.concept.suffix)],
```
doctrino, line 569: Should we raise is `concept.instance_souce` is not set? Is it safe to use `namespace[str(concept.concept.suffix)]` as a fallback?

nikokaoja: We need to support both pure generation of shacl without relation to instances and when those instances are present.

doctrino (dismissed):

PR #1308, opened 2025-10-01
[THIS-756] 👮 View tests

doctrino:
# Description

**Context**: We are changing the internal structure of the neat data models. We are dropping the physical data model format in Neat, and instead will have a one-to-one DMS format identical to the API. In principle, we could use the data classes that are in the `cognite-sdk`, but we decided to reimplement them in `pydantic`. This is to 1. Get better validation. 2. Better performance on serialization (pydantic does serialization in rust), 3. Make them easy to extend and connect to the Neat Conceptual model.

This PR implements validation for views. 

I also introduce a deviation from the view API. I set `connectionType` on primary properties, this is to use discriminator on `connectionType` to select the correct property type. 

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite/neat/_data_model/models/dms/_views.py
+++ b/cognite/neat/_data_model/models/dms/_views.py
@@ -94,10 +94,7 @@ class ViewRequest(View):
     )
 
     @field_validator("properties", mode="after")
-    def validate_properties_identifier(
-        cls,
-        val: dict[str, ViewRequestProperty],
-    ) -> dict[str, ViewRequestProperty]:
+    def validate_properties_identifier(cls, val: dict[str, ViewRequestProperty]) -> dict[str, ViewRequestProperty]:
         """Validate properties Identifier"""
         return _validate_properties_keys(val)
 


tests/tests_unit_v1/test_data_model/test_models/test_dms/test_views.py
```diff
@@ -1,23 +1,254 @@
...
+            "string should have at least 1 character",
+            "In properties.invalidEdge.single_edge_connection.type.space string should have at least 1 character",
+        },
+        id="Empty reference fields.",
+    )
+
+
+class TestViewRequests:
+    @pytest.mark.parametrize("data,expected_errors", list(invalid_view_definition_test_cases()))
+    def test_invalid_definitions(self, data: dict[str, Any], expected_errors: set[str]) -> None:
```
frxstrem, line 250: I think there should also be tests for _valid_ definitions

PR #1326, opened 2025-10-06
[THIS-774] 🕵 Humanize-validation

doctrino:
# Description

**Context** We are in the process of reimplementing the core part of Neat. 

This PR is a a split from #1323. It update the humanize_validation with options to make it fit in a table context.

## Bump

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite/neat/_utils/validation.py
+++ b/cognite/neat/_utils/validation.py
@@ -1,23 +1,63 @@
+from collections.abc import Callable, Mapping
+from typing import Literal
+
 from pydantic import ValidationError
 from pydantic_core import ErrorDetails
 
 
-def humanize_validation_error(error: ValidationError) -> list[str]:
+def as_json_path(loc: tuple[str | int, ...]) -> str:
+    """Converts a location tuple to a JSON path.
+
+    Args:
+        loc: The location tuple to convert.
+
+    Returns:
+        A JSON path string.
+    """
+    if not loc:
+        return ""
+    # +1 to convert from 0-based to 1-based indexing
+    prefix = ""
+    if isinstance(loc[0], int):
+        prefix = "item "
+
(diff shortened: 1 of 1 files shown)


cognite/neat/_utils/validation.py
```diff
@@ -1,23 +1,63 @@
...
+
+    Args:
+        loc: The location tuple to convert.
+
+    Returns:
+        A JSON path string.
+    """
+    if not loc:
+        return ""
+    # +1 to convert from 0-based to 1-based indexing
```
sighol, line 19: Are you sure that 1-based indexing is the most natural? I think we use 0-based indexing in other APIs, The time series API returns 0-based indexes in its error messages.

cognite/neat/_utils/validation.py
```diff
@@ -1,23 +1,63 @@
...
+    prefix = ""
+    if isinstance(loc[0], int):
+        prefix = "item "
```
sighol, line 22: Why do you want a space after `item`?
I tried this: `as_json_path((1, "test", 3, "check", "some_prop"))`
And it returned: `'item [2].test[4].check.some_prop'`

tests/tests_unit/test_utils/test_validation.py
```diff
@@ -0,0 +1,79 @@
...
+                {"name": "Frank", "age": "fifty"},
+                {
+                    "parent_loc": ("Data", 2),
+                    "humanize_location": lambda loc: f"at position {loc[1]} in {loc[0]}",
+                    "field_name": "column",
+                    "field_renaming": {"age": "Years"},
+                },
+                {"In at position 2 in Data input should be a valid integer, unable to parse string as an integer"},
+                id="Combined custom parameters",
+            ),
```
sighol, line 69: as_json_path is not tested. Please add a test where `parent_loc` is set and `humanize_location` is not.

PR #1612, opened 2026-08-03
Add validation for reverse connections pointing through reverse prope…

poocog:
# Description

Detect when a reverse direct relation's through property is itself a reverse relation, including cyclic A↔B cases, with a clear error explaining the mapping never resolves to a container. De-duplicate REVERSE-003 for this case and add snapshot tests.

## Bump

- [x] Patch
- [ ] Skip

### Added/Changed/Improved/Removed/Fixed

- Add NEAT-DMS-CONNECTIONS-REVERSE-010 (ReverseConnectionThroughPropertyIsReverse) to detect when a reverse direct relation points through another reverse property instead of a direct relation that maps to a container.
- Report cyclic reverse connection paths (e.g. A↔B) in the error message to make large-model debugging easier.
- De-duplicate REVERSE-003 for reverse-through-reverse cases and add snapshot/integration tests.

### DIFF
--- a/cognite/neat/_data_model/rules/dms/__init__.py
+++ b/cognite/neat/_data_model/rules/dms/__init__.py
@@ -21,6 +21,7 @@ from ._connections import (
     ReverseConnectionSourceViewMissing,
     ReverseConnectionTargetMismatch,
     ReverseConnectionTargetMissing,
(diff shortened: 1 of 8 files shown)


poocog:
Superseeded by #1613

PR #1613, opened 2026-08-04
Improve REVERSE-003 for cyclic reverse relation properties

poocog:
# Description

Improve NEAT-DMS-CONNECTIONS-REVERSE-003 for cyclic reverse relation properties, and fix the issues UI so differentiated fix messages show per issue when multiple issues share the same rule code.

### REVERSE-003 validation

- Extend REVERSE-003 with a conditional error message when a reverse connection's `through` property is itself a reverse direct relation.
- Report cyclic reverse connection paths in the error message.
- Provide distinct suggested fix text for cyclic vs generic non-direct through cases.
- Keeps validation rule count at 38 (no separate REVERSE-010).

### Issues UI

- Show suggested fix under each issue in grouped views (all validation types, not only REVERSE-003).

### Analysis mappings

- Add `all_reverse_connections` for validators; filter reverse-through-reverse from `reverse_to_direct_mapping` and `connection_end_node_types`.

Manual verification: load `tests/data/snapshots/local/cyclic_reverse_only` and run `pytest tests/tests_unit/test_data_model/test_rules/dms/test_bidirectional_connection_validator.py`.

## Bump

- [x] Patch
- [ ] Skip

### Added

- Neat will now validate and warn with a more informative warning if you have a reverse direct relation targeting another reverse direct relation.


cognite/neat/_data_model/rules/dms/_connections.py
Magssch: Since you use the term "direct relation property" (which I agree we should be doing) above, could you also change this text at the same time to say "is not a direct relation property"? Instead of "Direct connection property"

cognite/neat/_data_model/rules/dms/_connections.py
```diff
@@ -269,6 +300,44 @@ def validate(self) -> list[ConsistencyError]:
 
         return errors
 
+    def _find_reverse_cycle_path(
```
Magssch: I'm wondering whether this added complexity is worth it, without this method we would still be able to say that View A has a RDR that itself points to another RDR in View B. The added benefit of this method is that you can also show tha you e.g. have created a more complex cycle like ViewA->ViewB->ViewC->ViewA all consisting of RDRs. I'm not sure how much added value that provides however since DMS will just reject any RDR pointing to another RDR, no matter whether it would have formed a cycle or not. IMO we could remove this method and the cycle_suffix part of the error message as I think the rest of the new message is informative enough, what do you think?

Magssch (approved):
LGTM!

PR #1615, opened 2026-08-13
Add Toolkit template variable substitution for toolkit YAML reads

poocog:
﻿# Description

Resolve Toolkit `{{ variable }}` placeholders when reading module YAML with `format="toolkit"`, using Toolkit config (`default.config.yaml`, env overlays, `cdf.toml`, and module overrides). Options are collated in `ToolkitReadOptions` and config is loaded once per directory; strategy is documented in `_toolkit_variables.py`.

## Bump

- [x] Patch
- [ ] Skip

### Added

- Add Toolkit template variable substitution for `format="toolkit"` YAML reads via `toolkit_env`, `toolkit_config`, and `toolkit_version`.

### DIFF
--- a/cognite/neat/_data_model/importers/_toolkit_variables.py
+++ b/cognite/neat/_data_model/importers/_toolkit_variables.py
@@ -13,8 +13,10 @@ Strategy (pipeline)
    Toolkit projects for the path being read.
 
 2. **Resolve config chain** — Merge configs in Toolkit order:
-   ``default.config.yaml`` (base) then an environment overlay
-   (``config.<env>.yaml``, explicit ``toolkit_config``, or ``cdf.toml`` hints).
+   ``default.config.yaml`` (project and/or module-local), then an environment overlay.
+   Overlay selection: explicit ``toolkit_config``, then ``config.<toolkit_env>.yaml``,
+   then ``cdf.toml`` ``default_config_yaml`` / ``default_env`` (e.g. sandbox),
+   then ``config.dev.yaml`` / ``config.sandbox.yaml``.
    Prefer the organization directory from ``cdf.toml`` when present
    (e.g. ``beyond-the-plant/config.dev.yaml``).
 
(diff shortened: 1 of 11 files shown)


cognite/neat/_data_model/importers/_toolkit_variables.py
```diff
@@ -0,0 +1,502 @@
...
+5. **Substitute** — Replace ``{{ name }}`` in YAML text *before* parsing.
+   Unresolved names raise ``FileReadException`` with available variables listed.
+
+Performance note
+----------------
+Load config once per directory import via :class:`ToolkitProjectContext`, then
+reuse the flattened map for every YAML file under that directory.
+"""
+
+from __future__ import annotations
```
doctrino, line 40: This is for Python 3.9 and earlier which we have never supported for neat.

cognite/neat/_session/_physical.py
```diff
@@ -136,8 +145,9 @@ def _yaml(
         self,
         io: Any,
         format: Literal["neat", "toolkit"] = "neat",
-        data_model_file: Path | None = None,
+        data_model_file: Path | str | None = None,
         fix: bool = False,
+        options: ToolkitReadOptions | None = None,
```
doctrino, line 150: Ideally, we want to make the public API, which this is part of only basic types (string, integers, boolean). Here you are introducing a new class that the user must import in their notebook. This class is not publicly exposed to the user, so they cannot import it. 

Looking at `ToolkitReadOptions`, I see that it might be necessary, but I am not really sure. I think this needs some pondering. 

Current implementation looks like this
```python

from cognite.neat._data_model.importers._toolkit_variables import ToolkitReadOptions
neat.physical_data_model.read.yaml(io="path-to-toolkit-directory", format="toolkit", options=ToolkitReadOptions(...))
```
This is starting to feel clunky and unintuitive. 

An alternative is to split this into two methods, one for reading neat_yaml and one for toolkit_yaml


```python
neat.physical_data_model.read.yaml_toolkit(io="path-to-toolkit-directory", format="toolkit", env="dev")
```
and 

```python
neat.physical_data_model.read.yaml_neat(io="path-to-neat-yaml", format="toolkit", fix=True)
```

Then the toolkit version is for running a CI, while the neat version is for developing. However, that raises more questions like should we support neat format in CI through an interface that is intended to be used in a notebook?

doctrino (commented):
See my longer comment on the `ToolkitReadOptions`. 

I think this is a question of should we do this? And if so, how should we do this? Not the implementation itself. I am happy to take a discussion on it when I am back, or take it without me before then.

